PR #40, opened 2013-12-13
Get the NuGet Package Restore step to work with Mono

asbjornu:
This branch should hopefully get NuGet to work with Mono (on Linux), which will make Travis happy.


asbjornu:
Seems like most stuff is working on Mono now except GitVersionTask. Closing this.

PR #65, opened 2015-01-04
Moved projects around - Added PCL support & Other Frameworks

### LINKED ISSUE #60: PCL Support for MonoTouch
I'd love to see PCL support for MonoTouch so I can use this in my iOS applications. I'd be happy to put the PR in if you're in agreement and don't already have that work underway.

thedillonb:
# Readme First

The following adds PCL (Fixes #60) support via the Bait and Switch method. PCL exceptions are not verbose enough to appropriately fill out the SentryException object. Thus, it's easier to just switch the dll out for the non-PCL version and still retain all that juicy debug information.

I touched a lot of code, and please feel free to reject if you don't feel comfortable with the changes. I'd rather try to incorporate before forking into a similar project. 

Also, i suggest pulling this down instead of trying to look at it on GitHub

## What did I do?

* There now 4 main projects: SharpRaven (.net 4.5) SharpRaven_Monotouch (Monotouch), SharpRaven_Andorid (Android) and SharpRaven_Portable (PCL)
  * They all almost all share all the code, with different frameworks of course - except PCL.
  * Portable is all stubs. http://log.paulbetts.org/the-bait-and-switch-pcl-trick/ for more info if you've never seen it before.
* Dropped support for .NET 4.0 so TPL can be used.
  * If users are still on 4.0 (which by now is pretty old) I would just NuGet the older versions
  * async/await is more appropriate for the sending of data
* DTOs now have no logic in them - the filling of the data is relegated to the factory
  * This allows other frameworks to swap out the factory instead of ~4 DTO (transfer objects) objects. Not all frameworks have the same type of objects (System.Diagnostics.Stacktrace, Frame, etc...) 
* The `dynamic` stuff has been moved to a factory decorator since other frameworks, like MonoTouch or Android, don't support the DLR. Thus, the `HttpContextFactory` is not included in those projects, instead of having a bunch of `#if` `#else`
* Removed all the deprecated methods. Since I already broke the contract by adding TPL support I thought I might as well get rid of those.
* Renamed `ExceptionFrame` to `SentryExceptionFrame` per the comment somone had already made about consistency
* Removed a few unit tests. Since the sentry transfer objects are dumb, all the logic for filling out these objects is found in the `ISentryRequestFactory`. Thus, I made sure coverage was there on that code and removed anything else that double tested.
* Moved the projects down a level. Having everything in this `app` folder seemed like an extra folder that can be avoided.


asbjornu:
Thanks for this PR! I don't have time to look at it yet, but I just want you to know I appreciate the work you've put in and that I'll take a look as soon as time allows.

The only thing I can say I disagree with is the removal of the `app` folder. I quite like separating tests, tools, packages and the main application from each other. Having a folder in the root of `src` to distinguish this, is useful. Especially as we add more application-related projects to the solution, as you have done here.

thedillonb:
No worries. Take your time. I can definitely compromise on the app folder. I'll add it back and update the pull request in the next day or two.

asbjornu:
Brilliant. It might pass the TeamCity build with the `app` folder back too.

thedillonb:
I reworked the pull request with the original folder structure. I also added in support for unified Xamarin iOS.

asbjornu:
Thanks, @thedillonb, great work! But sadly, the build still fails. The errors are:
```
Z:\Builds\work\c42d5cd0fcf05c45\src\app\SharpRaven\SharpRaven_Android.csproj(88, 3): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Xamarin\Android\Xamarin.Android.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
Z:\Builds\work\c42d5cd0fcf05c45\src\app\SharpRaven\SharpRaven_Monotouch.csproj(84, 3): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Xamarin\iOS\Xamarin.MonoTouch.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
Z:\Builds\work\c42d5cd0fcf05c45\src\app\SharpRaven\SharpRaven_Xamarin.iOS.csproj(76, 3): error MSB4019: The imported project "C:\Program Files (x86)\MSBuild\Xamarin\iOS\Xamarin.iOS.CSharp.targets" was not found. Confirm that the path in the <Import> declaration is correct, and that the file exists on disk.
```
Can you please have a look at these errors and attempt to fix them so TeamCity is pleased? :smile:

GeirGrusom:
@asbjornu you need to install Xamarin on the build server. The build error is because Xamarin's MSBuild components are missing from the build server.

asbjornu:
@GeirGrusom As SharpRaven is using [CodeBetter's TeamCity Server](http://teamcity.codebetter.com/project.html?projectId=SentryRavenCsharp&tab=projectOverview), it's not up to me to install Xamarin. Is Xamarin required or can something be done to make it work without it?

GeirGrusom:
The command line build requires a business license, so I doubt it. I think the only solution then is that the bait-and-switch tactic is abandon in favor of a pure PCL.

This pull-request worked for me, and builds perfectly fine, with the exception of some hiccups with GitVersion, but I'll see if I can make a pull request without using bait-and-switch.

asbjornu:
@GeirGrusom Brilliant! Looking forward to the PR.

GeirGrusom:
I have created a pull request #70 which should implement this feature using reflection instead of bait-and-switch.

I think this pull request still has a lot of merit since it is a lot more capable of getting proper stack traces. Using reflection at least feels a little like fumbling in the dark.

asbjornu:
@thedillonb Can you have a look at #70 and see if it solves the problem in a satisfactory manner? If not, please provide details in #70 on how it can be improved, since it seems to be in a more merge-ready state than this PR.

asbjornu:
I'm closing this in favor of #70, since it's easier to merge than this.

PR #70, opened 2015-06-19
Add support for PCL Profile47

GeirGrusom:
This pull-request implements PCL support by using reflection and dynamic dispatch.

The biggest issue was getting sensible information from stack traces. It will try to dynamically create a StackTrace but if that fails it will build it using regular expressions, which is a less than ideal solution.

# Changes 

`SentryRequest` now tries to get`Exception.Source` using reflection, and if it fails, it will just default to an empty string.

`SentryStackTrace` has the most significant changes. I pulled the code to retrieve stack traces into a seperate method, and made three implementations: .NET Proper, PCL using dynamic and reflection, PCL using regular expressions. PCL should fall back to regular expressions if mscorlib does not contain a `System.Diagnostics.StackTrace` class. It will also fall back if it fails for any other reason, If regex fails it will simply do nothing. I don't think it is likely that the regex fails, but in any case it gives poor stack information so it is a very undesirable solution.


GeirGrusom:
I need to do some more device testing before I would deem this merge-ready.

asbjornu:
@GeirGrusom How is the device testing going?

GeirGrusom:
I have been fighting with GitVersion, Xamarin's NuGet package manager (the package itself works, but getting Xamarin to actually install it is more complicated) and the MSBuild scirpts.

I plan on getting this working this week, and I think I will squish the embarrasingly large amounts of commits and open a new pull request.

asbjornu:
@GeirGrusom Sounds good! Perhaps the new GitVersion works better than the old one? Looking forward to seeing that PR! :smile:

asbjornu:
@GeirGrusom It would be awesome to see a new go at this based on the latest `develop` branch. Just hit me with a new pull request whenever you're done with it. Closing this.

Issue #79, opened 2015-09-17
Setup Travis CI


dcramer:
http://docs.travis-ci.com/user/languages/csharp/

asbjornu:
This requires Mono support, which already fails in [TeamCity](http://teamcity.codebetter.com/viewType.html?buildTypeId=bt999), so I'm not sure what we can do to make this useful. And since we have a working TeamCity build (for Windows), what use will Travis have?

dcramer:
Mostly that Travis is much more accessible to people than TeamCity. I still want to make this work but haven't had a chance yet. It sounds like it simple comes down to getting Mono support working (which tbh we want to do anyways) so I'll keep this open as a point of reference for now.

asbjornu:
Sure. If you can find anyone with time and interest in fixing Mono compatibility, that would be great!

PR #80, opened 2015-09-17
Initial Travis CI config

dcramer:
Fixes GH-79


dcramer:
Doesnt work yet, but seems like it should be easy enough to get basics running.

Issue #86, opened 2015-11-13
Exception Data is not send to Sentry by default


jmalczak:
So i had to manually set Exception.Data property to be send as extra field. I believe it should be sent by default to sentry.

dcramer:
Not a C# dev myself, can you explain what Exception.Data contains? I want to make sure we support use cases fairly upstream.

jmalczak:
Sure. It's the same concept as extra field. This is taken from MSDN documentation "a collection of key/value pairs that provide additional user-defined information about the exception". You can read more about it here: https://msdn.microsoft.com/en-us/library/system.exception.data(v=vs.110).aspx

asbjornu:
@jmalczak It's pretty easy to just shove the `Data` from the first exception in the stack inside the `extra` dictionary, but what about `Data` in `InnerException`s? Should each `Data` dictionary have their keys prefixed with the name of the exception they occurred in and then flattened to one dictionary? And how should this be combined with the existing `extra` dictionary? How should collisions in key names be handled?

A pull request tackling these issues would be highly appreciated. :)

jmalczak:
Yeah if you think about inner exceptions it's not so trivial. Thx for the branch, I will take a look at it.

jmalczak:
Ok. So i have removed this test from Integration and created new ones in JsonPacketFactory tests, i think it is fine to have it there since Capture is just wrapper for passing arguments. I have also fixed compilation error and issue with UniqueKeys. Please take a look, give me some feedback. Regarding data itself, maybe we can prefix exception data somehow, for example when you invoke Capture like that 
```
CaptureException(exception, extra: new { key = "value" }) and exception.Data = { { "key2", "value2 } }
```
maybe extra can look like this:
```javascript
{
 key: "value",
 exceptionEXCEPTIONTYPEData: { key2: "value2" },
 exceptionINNEREXCEPTIONTYPEData: {}
}
```
and so on. What do you think? It would be easy to distinct exception data and extra since if you don't know implementation (you don't know that you add extra first and exception data later) it's not easy to figure out if key comes from data or extra if they are the same.

Commit with changes: https://github.com/jmalczak/raven-csharp/commit/bf275fdcb4f7eca730cdd9032cfa45f8e5ae06c9

asbjornu:
@jmalczak Thanks. I've merged and done a bit more on this. Let's continue the discussion in the #87. I'm closing this.

PR #87, opened 2015-11-18
Capture Exception.Data

asbjornu:
This is an attempt at resolving #86.

### DIFF
--- a/src/app/SharpRaven/Data/ExceptionData.cs
+++ b/src/app/SharpRaven/Data/ExceptionData.cs
@@ -31,6 +31,8 @@
 using System;
 using System.Collections.Generic;
 
+using Newtonsoft.Json;
+
 namespace SharpRaven.Data
 {
     /// <summary>
@@ -38,7 +40,7 @@ namespace SharpRaven.Data
     /// </summary>
     public class ExceptionData : Dictionary<string, object>
     {
-        private readonly Exception exception;
+        private readonly string exceptionType;
 
 
         /// <summary>Initializes a new instance of the <see cref="ExceptionData"/> class.</summary>
@@ -48,14 +50,14 @@ namespace SharpRaven.Data
             if (exception == null)
                 throw new ArgumentNullException("exception");
 
-            this.exception = exception;
+            this.exceptionType = exception.GetType().FullName;
 
             foreach (var k in exception.Data.Keys)
             {
                 try
                 {
                     var value = exception.Data[k];
-                    var key = JsonPacketFactory.UniqueKey(this, k);
+                    var key = k as string ?? k.ToString();
                     Add(key, value);
                 }
                 catch (Exception e)
@@ -72,13 +74,38 @@ namespace SharpRaven.Data
         }
 
 
-        /// <summary>Adds the current <see cref="ExceptionData"/> instance to the specified <paramref name="dictionary"/>.</summary>
-        /// <param name="dictionary">The dictionary to current <see cref="ExceptionData"/> instance to.</param>
-        public void AddTo(IDictionary<string, object> dictionary)
+        /// <summary>Gets the type of the exception.</summary>
+        /// <value>The type of the exception.</value>
+        [JsonProperty("type")]
+        public string ExceptionType
         {
-            var key = String.Concat(this.exception.GetType().Name, "Data");
-            key = JsonPacketFactory.UniqueKey(dictionary, key);
+            get { return this.exceptionType; }
+        }
+
+
+        private void AddTo(IDictionary<string, object> dictionary)
+        {
+            var key = String.Concat(ExceptionType, '.', "Data");
+            key = UniqueKey(dictionary, key);
             dictionary.Add(key, this);
+        }
+
+
+        private static string UniqueKey(IDictionary<string, object> dictionary, object key)
+        {
+            var stringKey = key as string ?? key.ToString();
+
+            if (!dictionary.ContainsKey(stringKey))
+                return stringKey;
+
+            for (var i = 0; i < 10000; i++)
+            {
+                var newKey = String.Concat(stringKey, i);
+                if (!dictionary.ContainsKey(newKey))
+                    return newKey;
+            }
+
+            throw new ArgumentException(String.Format("Unable to find a unique key for '{0}'.", stringKey), "key");
         }
     }
 }
--- a/src/app/SharpRaven/Data/IJsonPacketFactory.cs
+++ b/src/app/SharpRaven/Data/IJsonPacketFactory.cs
@@ -2,21 +2,21 @@
 
 // Copyright (c) 2014 The Sentry Team and individual contributors.
 // All rights reserved.
-//
+// 
 // Redistribution and use in source and binary forms, with or without modification, are permitted
 // provided that the following conditions are met:
-//
+// 
 //     1. Redistributions of source code must retain the above copyright notice, this list of
 //        conditions and the following disclaimer.
-//
+// 
 //     2. Redistributions in binary form must reproduce the above copyright notice, this list of
 //        conditions and the following disclaimer in the documentation and/or other materials
 //        provided with the distribution.
-//
+// 
 //     3. Neither the name of the Sentry nor the names of its contributors may be used to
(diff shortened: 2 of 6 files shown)


asbjornu:
@jmalczak It now captures nested exception data with a new `ExceptionData` class, which is built up hierarchical; one for each `.InnerException`. The hierarchy of `ExceptionData` is finally added to the `extra` dictionary by the name of the exception type plus "Data", which is first ensured to be unique. I think the design is good and that the tests prove that it works. Could you test how this looks in the Sentry GUI?

jmalczak:
@asbjornu Rally cool. Thanks a lot. This is how it looks like in Sentry. I have 2 exceptions Notimplemented which is inner and Exception which is main exception:

image

It would be perfect if we can just render ExceptionData property in sentry since all the other properties are really a bias.

asbjornu:
@jmalczak I think that looks good! But I'm not sure what you mean with this:

> It would be perfect if we can just render ExceptionData property in sentry since all the other properties are really a bias.

Could you please elaborate?

jmalczak:
Sure. It would be perfect if we can just have this part rendered to Sentry:
image

Other properties like IsFixedSize, IsReadOnly, IsSynchronized are really not needed.

asbjornu:
@jmalczak Hm. I'm not sure where those are coming from. The only place we're doing `.GetProperties()` is on [on line 162 in `JsonPacketFactory`](/getsentry/raven-csharp/blob/feature/exception-data/src/app/SharpRaven/Data/JsonPacketFactory.cs#L162) on the `extra` object, not on any `Exception` object. Do you have any idea where they are coming from?

asbjornu:
@jmalczak Ah, of course. `JsonPacket.Extra` is no longer an `object`, but an `IDictionary<string, object>` and that's where the other properties must be coming from.

jmalczak:
@asbjornu  i did a small fix, take a look https://github.com/jmalczak/raven-csharp/commit/a1fc8b062601e2879651ff15c8d5d7c9f29bcff4

jmalczak:
Then additionalData looks like that:

image

asbjornu:
@jmalczak I'd like to just get rid of the `Dictionary<string, object>` properties instead of serializing to `string`. Could you please help me figure out how to reproduce the serialization problem in a unit test, so we can make sure we fix the problem properly?

jmalczak:
Ok, I had some more time and investigated it further. So ignore my commit since the problem actually is in the code serializing extra param.

```
            if (extra != null)
            {
                foreach (var property in extra.GetType().GetProperties())
                {
                    try
                    {
                        var value = property.GetValue(extra, BindingFlags.Default, null, null, null);
                        var key = UniqueKey(result, property.Name);
                        result.Add(key, value);
                    }
                    catch (Exception e)
                    {
                        Console.WriteLine("ERROR: " + e);
                    }
                }
            }
```
You just iterate over properties, and before Json.NET was serializing value. So if I just set exception data, everything is fine. But if I set extra as well then we start having additional properties. I don't really know good solution to that. We could create new object:

```
new { Extra: oldExtra, ExceptionData: exceptionData }
```

and return it from Convert method. This way everything will be serialized by Json.NET so it will look ok. But this will break backwards compatibility, so if someone was setting extra: 
```
new { Test = "value" }}
```
he will think that Test property should be saved like that:

```
{
 test: "value"
}
```
but will be saved like that
```
{
  extra: { test : "value" }
}
```
We could also add some logic to get rid of those properties for Dictionary, but this won't be really backwards compatible as well since you won't be able to replicate exact serialization behavior of Json.NET.

I would go with this option if backwards compatibility is not an issue:
```
new { Extra: oldExtra, ExceptionData: exceptionData }
```
then you can simplify convert method to look like that:
```
        private static object Convert(object extra, Exception exception = null)
        {
            if (exception == null && extra == null)
            {
                return null;
            }

            return new { Extra = extra, ExceptionData = new ExceptionData(exception) };
        }
```

asbjornu:
@jmalczak Thanks for the investigation. But it's still not clear to me how this is happening. Are you able to reproduce in a test?

jmalczak:
Here is an example of failing test: https://github.com/jmalczak/raven-csharp/commit/a1f0787c27441e99060e932eda65356f96ee27cb

asbjornu:
@jmalczak Thanks. I've added another test and some code that should ensure that stuff like this is handled much better. How does it look to you?

jmalczak:
I don't think this is a good approach. You are writing your own serializer. And this is not such a straight forward thing. I would prefer just to do this:
```
 private static object Convert(object extra, Exception exception = null)
        {
            if (exception == null && extra == null)
            {
                return null;
            }

            return new { Extra = extra, ExceptionData = new ExceptionData(exception) };
        }
```

This way Dictionary will be serialized by Json.NET since it's used to serialize JsonPacket object.

asbjornu:
@jmalczak While I don't agree I'm writing my own serializer, I also think this might be going too far and a bit too clever. But I don't really like breaking backwards compatibility either. I think I have an idea that might work. Let me sketch it out and let me know what you think.

jmalczak:
@asbjornu What you can do is to create Json.NET JObject from extra object and then add exceptionData property as in example:

```
            var extraObject = new {TestField = "testvalue", TestField2 = 12};
            var extra = new Dictionary<string, string> { { "extra", "value" } };

            var exceptionData = new Dictionary<string, string> { { "exceptionData", "value" } };

            var jo = JObject.FromObject(extra);
            jo.Add("exceptionData", JToken.FromObject(exceptionData));

            Console.WriteLine(JsonConvert.SerializeObject(jo));

            var jo2 = JObject.FromObject(extraObject);
            jo2.Add("exceptionData", JToken.FromObject(exceptionData));

            Console.WriteLine(JsonConvert.SerializeObject(jo2));
```
In this case when you serialize it with Json.NET as it's done before sending JsonPacket, serialized value won't contain any unnecessary properties.

asbjornu:
@jmalczak Thanks for the idea, but I've tried that and it will fail if `extra` is an `Array`. I've also went further and tried to do stuff like:

```c#
if (extra.GetType().IsArray)
{
    var array = JArray.FromObject(extra);
    ...
}
```

but it ends up either mucking up the serialization with stuff we don't really want there (just like `Count` and such) or more reflection work to make the serialization prettier.

If you can figure out a way to serialize this that will pass the tests and make the serialization only contain what we want, I'd be grateful. If not, I think we need to just settle on an imperfect serialization although that doesn't sit well with me.

asbjornu:
This seems to do what we want, but it borders on "custom serialization" as you've mentioned on earlier attempts, @jmalczak:

```c#
private static object Merge(object extra, Exception exception = null)
{
    if (exception == null && extra == null)
        return null;

    if (extra != null && exception == null)
        return extra;

    if (extra == null)
        return new ExceptionData(exception);

    JObject jo;

    if (extra.GetType().IsArray)
    {
        jo = new JObject();
        var array = JArray.FromObject(extra);

        if (array.All(x => x is JObject))
        {
            var properties = array.Cast<JObject>().SelectMany(o => o.Properties()).ToArray();

            for (var i = 0; i < properties.Length; i++)
            {
                var property = properties[i];
                JProperty nextProperty;
                if (property.Name != "Key" || properties.Length <= i || (nextProperty = properties[i + 1]).Name != "Value")
                    continue;

                var key = property.Value.ToString();
                var value = nextProperty.Value;
                jo.Add(key, value);
            }
        }
    }
    else
        jo = JObject.FromObject(extra);

    jo.Add("ExceptionData", JToken.FromObject(exception.Data));

    return jo;
}
```

asbjornu:
@jmalczak I think I have a solution now with a lot less custom code, basing it instead on `JContainer.Merge()` to glue the objects together. I'll push when I've cleaned it up and fixed the tests.

asbjornu:
@jmalczak Please have a look at what I've arrived at now. I think it achieves the serialization we want without incurring to much custom serialization logic. There's just a bit covering arrays of `KeyValuePair<,>`, but it is very specific and shouldn't cause anything to fail as far as I can see. If you can come up with test cases that might break the current logic, please let me know.

PR #89, opened 2015-12-10
Better logscrubber docs

timorzadir:
If documentation exists, it should be correct.


dcramer:
@timorzadir was confused about SocialSecurityFilter until I looked at the file:

https://github.com/timorzadir/raven-csharp/blob/better-logscrubber-docs/src/app/SharpRaven/Logging/Filters/SocialSecurityFilter.cs

Thanks!

dcramer:
Quick aside, do you mind fixing the tests? Looks like something is asserting the filters are present.

timorzadir:
My mindset was set on fixing the documentation, so I missed the tests. Sorry about that. :)

PR #90, opened 2016-01-05
Remove GitVersionTask

dcramer:
This is currently incompatible with Mono. Supposed to be fixed in 4.0.0 but there's not yet a release available.

https://github.com/GitTools/GitVersion/issues/264


dcramer:
Unless anyone has a good reason why we need this right now I'm going to kill it since its effectively a blocker for cross platform support

asbjornu:
I'd appreciate time to comment before something like this is merged. GitVersionTask is used to version the project, both at the assembly level and for the published NuGet package. Without it, releasing both unstable and stable builds of SharpRaven becomes a hell of a lot harder.

A feature branch targeting Mono would be OK, but ignoring the "develop" branch and merging this into master directly is not how I would prefer to have it handled, to put it mildly.

The way I see it, easy versioning and release management is a lot more important than Mono support at the moment. I understand the wish to have a working Travis build, but we do have a TeamCity build that works so I don't understand the desperate need to rush it.

Please revert this.

dcramer:
The library fully works under Mono with this change.

asbjornu:
I appreciate that, but respectfully think that having simple version and release management is more important than Mono support. That is; Mono support can wait until GitVersionTask 4.0 is released. In my humble opinion.

PR #93, opened 2016-01-05
Support default tags on client

dcramer:


src/tests/SharpRaven.UnitTests/RavenClientTests.cs
```diff
@@ -161,7 +162,6 @@ public void Release_IsNullByDefault()
         [Test]
         public void CaptureMessage_SendsLogger()
         {
-            var project = Guid.NewGuid().ToString();
```
dcramer: bunch of unused code from previous patches

asbjornu:
Please use develop as base and merge target for features. I'd also love to be somewhat involved before stuff like this is merged, as I've been the project's main maintainer for a while and  feel I know C# and its community rather well.

dcramer:
@asbjornu not making any massive changes right now, just trying to expand API support and test coverage (with a goal to better support current spec and cross platform)

That said, digging into this, it seemsl ike there might have been other PRs merged into master, though I'll dig into the git history

PR #95, opened 2016-01-06
Merge develop

dcramer:
This merges in a number of commits made against master (eek!) instead of develop.

It reverts removal of GitVersionTask pending further discussion.


dcramer:
Ugh wrong branch

Issue #94, opened 2016-01-06
Move towards an Event Builder


dcramer:
The current way that events are generated makes it tough to handle a lot of arbitrary attribute (due to C#'s lack of something like kwargs).

For example, every time a new arg is added, it gets added to CaptureException, and thus to JSONPacketFactory.Create.

It'd make sense to take the unique bits (exception) and split them into their own generative items, possibly something akin to:

```
event = BuildEvent()
event.tags = ...
client.Capture(event)
```

asbjornu:
I agree that this would lead to cleaner extensibility. Adding an overload to `CaptureException`, or perhaps even an extension method, will preserve backwards compatibility too. :+1:

asbjornu:
@dcramer I've created #98 that we can play around with until we reach a design we're happy with. I'm already quite happy with how it looks.

PR #96, opened 2016-01-06
Various changes (reverted from `master`)

dcramer:
This merges in a number of commits made against master (eek!) instead of develop.

The pending item is to remove GitVersionTask as we need to support Mono. We can either work around whatever it does today, or find a way to use the unreleased 4.0.

- Add Environment
- Add Travis CI support
- Add default Tags
- Ensure Release is set on packets


dcramer:
Ok so the main thing is dealing with GitVersionTask here.

The only other major changes are:

- We should figure out #94
- Removed Dsn from Send() as it's available to the class

dcramer:
Also I'll force push to master to revert changes there

asbjornu:
> Add Environment

Absolutely.

> Add Travis CI support

I too want this, but I'd prefer a `mono` feature branch to cover it, so we can keep the project in working order while slowly getting everything up and running on Mono. I understand that GitVersionTask is what's keeping us from doing that at the moment, but I really want to stick with it to simplify version and release management.

> Add default Tags

Good.

> Ensure Release is set on packets

Good.

> We should figure out #94

Absolutely. It's a good idea.

> Removed Dsn from Send() as it's available to the class

Yep.

> Let's change default branch to develop

Yes please.

> or switch away from classical GitFlow (IMHO it provides no advantages, and we dont use it elsewhere in the getsentry org)

We use GitFlow in SharpRaven because [GitVersion](/GitTools/GitVersion) bases its versioning scheme on GitFlow or GitHubFlow. I use the `develop` branch to release [unstable packages of SharpRaven](https://www.nuget.org/packages/SharpRaven/1.5.1-unstable0134), which is impossible without a separate branch.

Using the rest of GitFlow with `feature/`, `hotfix/`, etc. just makes it easier to understand what each branch is and why it exists. So I don't understand why we should switch away from GitFlow and what advantages not using it would give us.

asbjornu:
Also, regarding the failing tests; the `LastEvent` being used in `RavenClientTests` doesn't exist. What's it supposed to do?

dcramer:
Do you know if there is a way we could change GitVersionTask to conditionally load? either not load in tests, or only load for the release processes. Maybe that's as simple as adding another solution?

dcramer:
My limited C# exposure failed me -- couldn't figure out how to mark Tags as readonly. Probably not a big deal?

dcramer:
I went ahead and updated the default branch to be ``develop``

asbjornu:
> My limited C# exposure failed me -- couldn't figure out how to mark Tags as readonly. Probably not a big deal?

Every detail matters, imho. It's a low cost on our part to make the usage of the client more pleasant and with fewer surprises and rough edges. I can take care of it.

> I went ahead and updated the default branch to be `develop`

Excellent.

PR #98, opened 2016-01-14
Added SentryEvent

asbjornu:
- Added SentryEvent class.
- Added Capture() and CaptureAsync() methods on IRavenClient taking SentryEvent.
- Obsoleted the CaptureMessage() and CaptureException() methods and their async counterparts on SentryEvent.
- Refactored all code to using the new SentryEvent class as its main code path.
- Added a couple of tests.


dcramer:
:+1:

Issue #99, opened 2016-01-21
SharpRaven messages not getting to GetSentry for some requests


uber-rob:
I'm using Log4net / SharpRaven / SentryAppender in a Web.API project with a custom IExceptionLogger set up to send exceptions to Sentry.

Locally, when testing with Postman, exceptions are thrown and they appear in Sentry a few seconds later. On production, when testing exceptions from a browser window the same happens.

However when a different internal server uses the API exceptions just disappear:
- log4net has no configuration issues and nothing in the trace log.
- Exceptions are being caught and e-mailed to me, so I know the IExceptionLogger is working.
- To rule out log4net or SentryAppender bug I have temporarily added direct exception reporting with the RavenClient and the behaviour is the same. Local OK, production OK when serving a local request, nothing from production when serving a remote request.

I'm using v1.5

I'm going to continue to investigate (for e.g. updating Newtonsoft.JSON, trying a SharpRaven pre-release package, logging something that's not an exception) but I thought I'd open this bug in case there was something else I could do to investigate.

uber-rob:
I'll also use the ErrorOnCapture property and report back if that catches anything (hadn't seen that in the documentation, sorry)

uber-rob:
OK I think I know what this is, and it's probably not SharpRaven. Some of my requests need to alter the way certificates are trusted manually (because of a really dodgy service underneath it) and that may be destroying the secure connections to GetSentry. Why this isn't happening locally is odd, but believably an environment concern.

uber-rob:
I've documented additional logging options to help future devs, hope this helps.

Amyyum:
Hi, may be you know the reason of exceptions are notified in sentry but don't e-mail to me ? Thanks very much!

PR #100, opened 2016-01-21
Documenting the ErrorOnCapture property of SharpRaven.

uber-rob:
Thought this might be helpful? I would not have opened https://github.com/getsentry/raven-csharp/issues/99 with this info more easily available.


asbjornu:
Much appreciated, @uber-rob! Thanks! :)

Issue #101, opened 2016-01-21
NRE in ExceptionFrame.cs (or maybe a mono bug)


razzmatazz:
Hi,

I have an issue where the following stack trace is emitted:
```
[ERROR] System.NullReferenceException: Object reference not set to an instance of an object
  at SharpRaven.Data.ExceptionFrame..ctor (System.Diagnostics.StackFrame frame) [0x00034] in /Users/bob/src/raven-csharp/src/app/SharpRaven/Data/ExceptionFrame.cs:64 
  at SharpRaven.Data.SentryStacktrace..ctor (System.Exception exception) [0x00043] in /Users/bob/src/raven-csharp/src/app/SharpRaven/Data/SentryStacktrace.cs:66 
  at SharpRaven.Data.SentryException..ctor (System.Exception exception) [0x00042] in /Users/bob/src/raven-csharp/src/app/SharpRaven/Data/SentryException.cs:60 
  at SharpRaven.Data.JsonPacket..ctor (System.String project, System.Exception exception) [0x00092] in /Users/bob/src/raven-csharp/src/app/SharpRaven/Data/JsonPacket.cs:77 
  at SharpRaven.Data.JsonPacketFactory.Create (System.String project, System.Exception exception, SharpRaven.Data.SentryMessage message, ErrorLevel level, IDictionary`2 tags, System.Object extra) [0x00000] in /Users/bob/src/raven-csharp/src/app/SharpRaven/Data/JsonPacketFactory.cs:101 
  at SharpRaven.RavenClient.CaptureException (System.Exception exception, SharpRaven.Data.SentryMessage message, ErrorLevel level, IDictionary`2 tags, System.Object extra) [0x00000] in /Users/bob/src/raven-csharp/src/app/SharpRaven/RavenClient.cs:153 
```

The problem is that code on line 64
```
62            var method = frame.GetMethod();
63            Filename = frame.GetFileName();
64            Module = (method.DeclaringType != null) ? method.DeclaringType.FullName : null;
```
is expecting for the `method` variable to be non-null as it is dereferencing `method.DeclaringType` immediately. In my case, however, this assumption fails for some reason and then I get a NRE.

The platform is mono 4.2.1 x86, raven-csharp @ master (commit 74a771068ef660558c3a1dd6ff75be21c1bcc45c).

[...]

PR #103, opened 2016-01-22
Bugfix nre in exceptionframe cs

razzmatazz:
Fix for https://github.com/getsentry/raven-csharp/issues/101, rebased on `develop`


asbjornu:
@razzmatazz There we go. Thanks! :)

PR #104, opened 2016-01-26
Replace MSBuild with Cake

asbjornu:
[Cake](http://cakebuild.net/) is a much nicer build system than MSBuild. This PR replaces MSBuild and custom batch script with a nice Cake script.


asbjornu:
Duplicate of #189.

Issue #111, opened 2016-03-16
Unable to set tags in 2.0


Meilu:
Just downloaded the new 2.0 version and it looks like i am no longer able to set the tags in the new Capture method. The old CaptureEvent does support a dictionary with tags as parameter but seems to be tagged as obsolete. How are we supposed to set tags in the new version?

rustyswayne:
I encountered the same issue when trying to set the ErrorLevel when passing a SentryEvent to the Capture method.  What is happening is the JsonPacketFactory instantiates the default type but does not assign all of the properties from the SentryEvent.

I just fixed it in my implementation and will submit a pull for to be reviewed.

asbjornu:
@Meilu Thanks for reporting this. I've released a [2.1 prerelease on NuGet](https://www.nuget.org/packages/SharpRaven/2.1.0-unstable0020) that I would very much appreciate if you could test and report back whether fixes the problem or not.

Meilu:
Hello guys, thank you for looking into this. I have tried the new version except i am still having the same issue. I checked out the source and i think the problem here is that the tags property on the sentryevent object is private without a set method. How can i pass a sentryevent with tags when it is set to private and without a set method?

Meilu:
Scratch my last comment, i didn't realise i could just call the add method on the dictionary.This issue is solved.

asbjornu:
@Meilu Awesome, thanks for reporting back! :smiley:

PR #112, opened 2016-03-16
Issue #111 - Addresses issue with JsonPacket not being fully built by JsonPacketFactory when passing SentryEvent

rustyswayne:
Fixes issue #111 by calling by determining what type of SentryEvent is being passed as a parameter (Exception or Message) and calling other existing factory methods to fully build the JsonPacket model.


asbjornu:
@rustyswayne Thanks!

PR #117, opened 2016-03-30
Log level is not set on JsonPacket

xpicio:
Hello, i was trying the new client with version 2.1.0-unstable0020 and every log message is logged as Error, it's seems you missed to set the ``Level`` property on new ``JsonPacket(string project, SentryEvent @event)`` constructor.


asbjornu:
@xpicio Excellent find! Thanks.

PR #121, opened 2016-04-19
SentryRequest.Data is empty

xpicio:
Hello, when the current HttpRequest is not an application/x-www-form-urlencoded the SentryRequest.Data will be empty, and the POST data will not be logged. I tried the attached patch on WebAPI and it seems working.

### DIFF
--- a/src/app/SharpRaven/Data/SentryRequestFactory.cs
+++ b/src/app/SharpRaven/Data/SentryRequestFactory.cs
@@ -36,6 +36,8 @@ using System.Linq;
 using System.Reflection;
 
 using Newtonsoft.Json;
+using System.IO;
+using System.Text;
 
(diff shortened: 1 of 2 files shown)


src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -119,6 +121,38 @@ public virtual SentryRequest OnCreate(SentryRequest request)
...
+            if (!HasHttpContext)
+                return null;
+
+            string body = String.Empty;
+
+            try
+            {
+                if (HttpContext.Request.Form.Count > 0)
+                {
+                    body = HttpContext.Request.Form;
```
asbjornu: That should probably use the `Convert()` function, no?

PR #122, opened 2016-04-20
Improved SentryRequest.Data log

xpicio:
Hello, i'm working on an improved solution to log in a better way the body content of the current request. Log the body content as string is not so great. I still need to do more test on MVC and Nancy, but it seems work on WebAPI, i will do it on the next days.

What do you think about this solution ? Do you have any suggestions ? Could we do better ?

A screenshot of a json POST request from a sentry 8 setup:
sentry-8-body


asbjornu:
@xpicio This looks like a neat solution. Give me a little time to look over it and I'm sure this can be merged. My first, initial thought, is: This needs testing. Think you can come up with a few tests that go through every media type and test its behaviour?

xpicio:
@asbjornu yes, i can do some test for every media type. Give me a few days.

asbjornu:
@xpicio Brilliant! 👍

asbjornu:
@xpicio How are those tests coming along?

xpicio:
@asbjornu really sorry for delay. At the moment i am really busy :(, and when i checked the HttpSimulator class it's seems there is no support to POST request through http body stream. HttpSimulator support only POST request with application/x-www-form-urlencoded, so the work seems more complex than expected :(

I think I'll be able to try an attempt next week.

asbjornu:
@xpicio I'm merging this in #138 with a couple of rudimentary tests. Please have a look at it and add more tests if you find time. Thanks for the nice feature!

PR #124, opened 2016-06-01
Update to fix issue with user and HttpContext

Sdhalm:
Issue with the way HttpContexts were gotten. Fixes ability to get actual context in system. Does not fix thread safe issue.


src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -1,271 +1,288 @@
...
+        {
+            // [Meilu] Since reflection is performance heavy, check if we have the httpcontext only once.
+            if (!CheckedForHttpContext)
+            {
+                TryGetHttpContextPropertyFromAppDomain();
+                CheckedForHttpContext = true;
+            }
+        }
+    }
 }
```
asbjornu: It's not easy doing a review when the entire file has been changed.

asbjornu:
@Sdhalm Thanks for this! But can you please try to fix the broken tests and not modify the entire file, so reviewing becomes a bit easier? 😄

Sdhalm:
Just curious from an understanding point of view is there a reason why you allow for the setting of the HttpContext. From what I can see in the code package I believe this should only be a get.

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -1,271 +1,288 @@
...
+namespace SharpRaven.Data
+{
+    /// <summary>
+    /// A default implementation of <see cref="ISentryRequestFactory"/>. Override the <see cref="OnCreate"/>
+    /// method to adjust the values of the <see cref="SentryRequest"/> before it is sent to Sentry.
+    /// </summary>
+    public class SentryRequestFactory : ISentryRequestFactory
+    {
+        [JsonIgnore]
+        internal static bool HasHttpContext
```
Sdhalm: Change this over to perform a GetHttpContext in the event we don't have the context already.

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -1,271 +1,288 @@
...
+        /// <summary>
+        /// Gets or sets the HTTP context.
+        /// </summary>
+        /// <value>
+        /// The HTTP context.
+        /// </value>
+        internal static dynamic HttpContext { get
+			{
+				try
+				{
```
Sdhalm: Changed this to directly use the property we got from reflection instead of the auto property from before

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -1,271 +1,288 @@
...
+                Console.WriteLine(exception);
+            }
+
+            return dictionary;
+        }
+
+
+        private static void TryGetHttpContextPropertyFromAppDomain()
+        {
+			if (HttpContext != null)
```
Sdhalm: changed this to not use the property as this would cause an infinite loop/out of memory exception is doing so.

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -1,271 +1,288 @@
...
+        }
+
+        private static void GetHttpContext()
+        {
+            // [Meilu] Since reflection is performance heavy, check if we have the httpcontext only once.
+            if (!CheckedForHttpContext)
+            {
+                TryGetHttpContextPropertyFromAppDomain();
+                CheckedForHttpContext = true;
+            }
```
Sdhalm: Removed additional setting code and placed in needed respective locations in code, see other comments.

Sdhalm:
fixed the build issue related to the unit test. Although again this really shouldn't get a set at all based on the changes I had made. I also added comments to the files

asbjornu:
@Sdhalm Why did you close the issue?

Sdhalm:
I have a better update for it that I am working on. Including the unit tests.
On Jun 6, 2016 3:42 AM, Asbjørn Ulsberg <notifications@github.com> wrote:@Sdhalm Why did you close the issue?

asbjornu:
@Sdhalm Ok, excellent. 👍  Looking forward to the pull request! 😄

PR #132, opened 2016-07-07
add net core support

ardymalihi:
Add .Net Core Support
Add Async
Removed Trace
Does not support compression
Changed Environment.UserName to Machine Name (Not Available)

---

This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/132)


src/app/SharpRavenNetCore/Data/ErrorLevel.cs
```diff
@@ -0,0 +1,68 @@
+﻿#region License
```
asbjornu: Shouldn't this and all other files be added as a link instead of copied?

asbjornu:
@ardymalihi Thanks for this! First, I think most of the files you've copied should be linked instead. Second, I wonder how this works alongside the existing packages? Is it just to bundle up the .NET Core `.dll` into the `.nupkg` just like we already do for .NET 4 and 4.5?

ardymalihi:
Agree. I was thinking of removing extra classes and filters and log levels as Asp.net Core is using its own logger provider.
It’s good to separate this project from the 4.x as it’s a cross platform

0xdeafcafe:
Any progress on the port? Would love to migrate over to this from Application Insights.

ardymalihi:
Hi,
I didn’t have a chance to work on it. I hope in a close future

trurl123:
Why did you add separate folder SharpRavenNetCore?? I compiled for .net standard without any problems. i just added target framework and upgraded Newtonsoft.Json.

ardymalihi:
It was not compiling at the time I created a separate branch

PR #134, opened 2016-08-10
Make sure User is initialized

### LINKED ISSUE #123: User is not populated properly in some cases
We noticed that in some cases the user passed to GetSentry is not populated correctly in GetSentry  

Looking at 
https://github.com/getsentry/raven-csharp/blob/develop/src/app/SharpRaven/RavenClient.cs#L288

```csharp
 packet.User = packet.User ?? this.sentryUserFactory.Create();
 packet.Request = packet.Request ?? this.sentryRequestFactory.Create();
```

This tweak is fixing the issue 

```csharp
packet.Request = packet.Request ?? this.sentryRequestFactory.Create();
packet.User = packet.User ?? this.sentryUserFactory.Create();
```
As SentryUser Create() method rely on SentryRequestFactory

```csharp
  public SentryUser Create()
        {
            SentryUser user;
            if (!SentryRequestFactory.HasHttpContext)
                user = new SentryUser(Environment.UserName);
            else
            {
                user = new SentryUser(GetPrincipal())
                {
                    IpAddress = GetIpAddress()
                };
            }

            return OnCreate(user);
        }
```

genne:
- Avoid storing current HttpContext statically (requires things to
happen in a certain order, and doesn't work with parallel requests)
- Updating test case to initialize the package in the same order as the
code does

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/134)

Fixes #123.

### DIFF
--- a/src/app/SharpRaven/Data/SentryRequestFactory.cs
+++ b/src/app/SharpRaven/Data/SentryRequestFactory.cs
@@ -31,7 +31,6 @@
 using System;
 using System.Collections.Generic;
 using System.Collections.Specialized;
-using System.ComponentModel;
 using System.Linq;
 using System.Reflection;
 
@@ -53,17 +52,33 @@ namespace SharpRaven.Data
             get { return HttpContext != null; }
         }
 
-        [JsonIgnore]
-        internal static bool CheckedForHttpContext { get; set; }
-
         /// <summary>
         /// Gets or sets the HTTP context.
         /// </summary>
         /// <value>
         /// The HTTP context.
         /// </value>
-        internal static dynamic HttpContext { get; set; }
+        internal static dynamic HttpContext
+        {
+            get
+            {
+	            TryGetHttpContextPropertyFromAppDomain();
+				
+                // [Meilu] If the currentHttpcontext property is not available we couldnt retrieve it, dont continue
+                if (!HasCurrentHttpContextProperty)
+                    return null;
 
+                try
+                {
+                    return CurrentHttpContextProperty.GetValue(null, null);
+                }
+                catch (Exception exception)
+                {
+                    Console.WriteLine("An error occurred while retrieving the current HTTP context: {0}", exception);
+                    return null;
+                }
+            }
+        }
 
         /// <summary>
         /// Gets or sets the CurrentHttpContextProperty
@@ -79,6 +94,9 @@ namespace SharpRaven.Data
             get { return CurrentHttpContextProperty != null; }
         }
 
+        [JsonIgnore]
+        internal static bool CheckedForHttpContextProperty { get; set; }
+
         /// <summary>
         /// Creates a new instance of <see cref="SentryRequest"/>
         /// for the current packet.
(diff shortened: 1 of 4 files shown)


asbjornu:
@genne Thanks for this! :heart: Do you think you can tackle the "multiple simultaneous requests" issue you pointed out in a test as well?

genne:
Absolutely, included a test case for it. Have verified it fails when the fix isn't there.

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -53,17 +52,33 @@ internal static bool HasHttpContext
...
         /// </summary>
         /// <value>
         /// The HTTP context.
         /// </value>
-        internal static dynamic HttpContext { get; set; }
+        internal static dynamic HttpContext
+        {
+            get
+            {
+	            TryGetHttpContextPropertyFromAppDomain();
```
asbjornu: Can you get rid of the extra indentation here?

src/tests/SharpRaven.UnitTests/Data/JsonPacketTests.cs
```diff
@@ -257,12 +239,27 @@ public void Constructor_WithHttpContext_UsertIsNotNull()
             });
         }
 
+        [Test]
+        [Category("NoMono")]
+        public void Constructor_WithHttpContext_UsertIsNotNull_Threaded()
```
asbjornu: Remove `t` in `Usert` :smile:

src/tests/SharpRaven.UnitTests/Utilities/HttpSimulator.cs
```diff
@@ -829,5 +830,10 @@ public void RemoveAt(int index)
         }
 
         #endregion
+
+        public void SetUser(string username)
```
asbjornu: I would prefer to have this as an extension method, if possible, since the `HttpSimulator` class is copied from god knows where. I can't remember. :smiley:

src/tests/SharpRaven.UnitTests/Data/JsonPacketTests.cs
genne: Was just copied from the test case above... ;)

genne:
Since there's a bunch of other HttpContext related utility methods available on the `HttpSimulator` I would prefer it you could set the current user when creating the request, but since it doesn't really depend on the simulator (it just uses `HttpContext.Current`) I simply moved it to the test case.

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -79,16 +94,16 @@ internal static bool HasCurrentHttpContextProperty
             get { return CurrentHttpContextProperty != null; }
         }
 
+        [JsonIgnore]
+        internal static bool CheckedForHttpContextProperty { get; set; }
```
asbjornu: Does this need to be internal? If not, I'd prefer it as a private field.

genne: It was just moved and renamed (from 'CheckedForHttpContext') to make it more clear what it actually did.

But totally agree that it should be made private, but if you go that way there's a bunch of other things to look into as well... :smirk:

asbjornu: > but if you go that way there's a bunch of other things to look into as well... :smirk:

Go wild!

homer-simpson-crazy

genne: Probably a different PR :smile:

asbjornu: Ok, just this property now, then? :smiley:

asbjornu:
Merged in #136.

PR #143, opened 2016-08-17
Identify InApp stack-frames to improve stack-traces

marcomorain:
In Sentry, some stack frames can be marked as InApp, which means that
other stack frames are shown collapsed in the UI by default. This allows
the reader to skip over system frames quickly when reading traces.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/143)

### DIFF
--- a/src/app/SharpRaven/Data/ExceptionFrame.cs
+++ b/src/app/SharpRaven/Data/ExceptionFrame.cs
@@ -80,6 +80,7 @@ namespace SharpRaven.Data
             Filename = frame.GetFileName();
             LineNumber = lineNo;
             ColumnNumber = frame.GetFileColumnNumber();
+            InApp = !IsSystemModuleName(Module);
         }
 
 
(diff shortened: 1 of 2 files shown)


marcomorain:
## Showing Full stack traces

full


## Showing only In App stack frames

in_app_only

marcomorain:
I had some c#6 only code present - http://teamcity.codebetter.com/viewLog.html?buildId=225595&buildTypeId=bt1000&guest=1

Will fix now.

src/app/SharpRaven/Data/ExceptionFrame.cs
```diff
@@ -219,5 +223,11 @@ public override string ToString()
 
             return sb.ToString();
         }
+
+        private static bool IsSystemModuleName(string moduleName)
+        {
+            return !string.IsNullOrEmpty(moduleName) &&
+                moduleName.StartsWith("System.", System.StringComparison.Ordinal);
```
asbjornu: Perhaps include `Microsoft.` and possible other prefixes known to be external? This should probably be configurable in the client as well, so people can adjust this as they see fit.

asbjornu:
@marcomorain Thanks for this! Besides my one comment, I think this looks good!

src/app/SharpRaven/Data/ExceptionFrame.cs
marcomorain: Good point.

Are you OK with me just adding `Microsoft.` to the check for now, and coming back later to make it configurable if it's requested?

I guess to make it configurable we could pass in a ISet<string> of known internal prefixes? Or we could allow the user to pass in a regex or a Func<string, bool> that we use to filter the frames?

asbjornu: @marcomorain Configuring a `Func<string, bool>` sounds most flexible, yea. I'm ok with merging this without configurability for now.

marcomorain:
@asbjornu I've pushed aff8ec9766bfe9bb3d8f30b60a6c684eee185d9e which treats `Microsoft.`  as an internal package, too.

marcomorain:
Thanks for the feedback @asbjornu , and thanks for the project! We are using raven-csharp it in anger, and it's working great :+1:

asbjornu:
@marcomorain Thanks for the contribution! :heart:

marcomorain:
:tada:

PR #153, opened 2016-09-06
Stacktrace order

genne:
According to Sentry the most recent call should be presented first, but
for all C# exceptions it's presented in a reversed order. This fix makes
sure they are ordered correctly.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/153)


src/app/SharpRaven/Data/SentryStacktrace.cs
```diff
@@ -91,7 +84,7 @@ public override string ToString()
 
             StringBuilder sb = new StringBuilder();
 
-            foreach (var frame in Frames)
+            foreach (var frame in Frames.Reverse())
```
asbjornu: Is it necessary to do `Reverse()` twice (here and on line 60 above)?

marcomorain:
Thanks for addressing this! I noticed that the traces were strangely ordered in UI last week and I thought it was a UI bug. 

Are chained exceptions also ordered correctly?

src/app/SharpRaven/Data/SentryStacktrace.cs
genne: Depends on how we want the `ToString` implementation to behave. There is a unit test making sure it returns a string being equivalent to the `Exception.ToString()`, and in order for that to work we have to reverse the frames here. 

(Though, I'm not sure if that unit test adds any direct value, and also it doesn't work on computers with other language settings than English. Does for example not work on my computer (Swedish).).

genne:
Yes, chained exceptions should behave the same as the final exception (they all use this constructor).

src/app/SharpRaven/Data/SentryStacktrace.cs
asbjornu: Ah, I see. I remember writing that test, thinking it would be valuable having Sentry's object model directly represent and reflect that of `System.Exception`. Ideas on how to make that test more i18n friendly?

genne: I would probably just remove the test, since it's not important for this library that the resulting string is exactly the same as the output of an `Exception`. Or change the test to only make sure `ToString` returns a non-empty string (and doesn't crash).

asbjornu: How about asserting that the string contains the type names of the exceptions, but no assertions of the overall format of the string?

genne: That would probably work, you want me to go ahead and change that?

asbjornu: Yes, that would be great!

genne: Sorry, took some time, but test case has now been updated.

asbjornu: Excellent! Think you can add a little comment above the two `Reverse()` statements to explain why they are performed?

genne: Done!

asbjornu:
@genne Wonderful, thanks!

PR #155, opened 2016-09-14
Demangle the function names of async tasks

marcomorain:
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/155)

### DIFF
--- a/src/app/SharpRaven/Data/ExceptionFrame.cs
+++ b/src/app/SharpRaven/Data/ExceptionFrame.cs
@@ -33,7 +33,6 @@ using System.Diagnostics;
 using System.Text;
 
 using Newtonsoft.Json;
-using System.Text.RegularExpressions;
 
 namespace SharpRaven.Data
 {
@@ -82,7 +81,6 @@ namespace SharpRaven.Data
             LineNumber = lineNo;
             ColumnNumber = frame.GetFileColumnNumber();
             InApp = !IsSystemModuleName(Module);
-            Demangle();
         }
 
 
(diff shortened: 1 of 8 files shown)


src/app/SharpRaven/Data/ExceptionFrame.cs
```diff
@@ -230,5 +232,28 @@ private static bool IsSystemModuleName(string moduleName)
                 (moduleName.StartsWith("System.",    System.StringComparison.Ordinal) ||
                  moduleName.StartsWith("Microsoft.", System.StringComparison.Ordinal));
         }
+
+        /// <summary>
+        /// Demangle function names create when C# compilers lower newer language features such as async/await calls.
```
marcomorain: typo - create -> created

src/tests/SharpRaven.WebTest/Default.aspx
```diff
@@ -1,20 +1,27 @@
-﻿<%@ Page Language="C#" %>
+﻿<%@ Page Language="C#" Async="true" %>
```
marcomorain: I will revert this file and write some proper tests.

asbjornu:
@marcomorain Thanks for this! But I think you've discovered a design flaw of the `ExceptionFrame` class. Having to invoke `Demangle()` explicitly after setting `Module` or `Function` is a bit surprising. Since the `Module` and `Function` properties have public setters, the `Demangle()` method should preferably be invoked on the setter's `value`. That sounds dirty, though, since the `Demangle()` method requires both `Module` and `Function` as input.

We could make the setters `private readonly`, but that would break backwards compatibility, although I doubt many people are creating these objects in their own code. I think we can get away with a minor version increment on such a change.

The only problem I see then is your test, which will then have to inject the necessary mock data for the `Function` and `Module` through the `StackFrame` object. I have no idea if that's possible at all.

Thoughts?

marcomorain:
The mutation of `ExceptionFrame` objects is not nice. I'm really open to changing the approach. 

Maybe a static function that takes an ExceptionFrame and returns a new, demangled one would be nicer? This would mean that there is less logic in the constructor, and no mutation.

asbjornu:
Hm. No, I think a breaking change with `private readonly` backing fields would be better.

marcomorain:
@asbjornu I have changed how I implemented this entirely. I have separated all demangling code out from creation of the stack traces themselves (no more unnecessary logic in the constructor).

The code code is now triggered from JsonPacket.cs after the SentryException objects is created, we then process them with a set of processors. There is a single processor at the moment, but the code is written to allow more. We had talked before about supporting Roslyn and the mono compiler.

There are 2 new functions - `StackTraceUtil.Demangle ` walks a stack trace and applies processors to all frames in the stack traces, and `StackTraceUtil.FixAsyncMoveNext` that fixes up the MoveNext function names. These two functions have been written to be testable separately.

I'll try to write an integration test shortly.

asbjornu:
@marcomorain: Got time to look at this?

marcomorain:
I'm pretty busy this week. I'll try to get a look at it tomorrow if I
can.



On Mon, 26 Sep 2016, at 16:03, Asbjørn Ulsberg wrote:
> @marcomorain[1]: Got time to look at this?
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/marcomorain
  2. https://github.com/getsentry/raven-csharp/pull/155#issuecomment-249596898

marcomorain:
Closing in favor of #164.

PR #156, opened 2016-09-14
Adding support for .Net 3.5 compilation

aldav82:
Currently, Raven CSharp works pretty well in most scenarios. However, there are still legacy .Net 3.5 projects out there that could get the benefits of Sentry. Currently these benefits are far from their reach, because the current version is only packaged for .Net 4.0 and above. These current changes will add the ability to package a .Net 3.5 version of the library inside the generated Nuget package.

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/156)


src/app/SharpRaven/Data/Requester.Net45.cs
```diff
@@ -30,13 +30,15 @@
...
+#if (!net35)
 using System.Threading.Tasks;
+#endif
 
 using Newtonsoft.Json;
 
 using SharpRaven.Utilities;
 
-#if !(net40)
+#if !(net40) && !(net35)
```
asbjornu: Perhaps a `#if (net45)` check would be neater?

src/app/SharpRaven/Data/Requester.cs
```diff
@@ -74,8 +75,7 @@ internal Requester(JsonPacket packet, RavenClient ravenClient)
             this.webRequest.Headers.Add("X-Sentry-Auth", PacketBuilder.CreateAuthenticationHeader(ravenClient.CurrentDsn));
             this.webRequest.UserAgent = PacketBuilder.UserAgent;
 
-            if (ravenClient.Compression)
-            {
+            if (ravenClient.Compression) {
```
asbjornu: I see you have a different code style than the project. Can you please not change the brace placement? 😃

src/app/SharpRaven/Data/SentryRequestFactory.cs
```diff
@@ -35,7 +35,9 @@
 using System.Linq;
 using System.Reflection;
 using System.Text;
-
+#if net35
+using System.Web;
```
asbjornu: I guess there's no simple way to avoid this `System.Web` reference?

src/app/SharpRaven/Data/Requester.Net45.cs
aldav82: I am not sure. Currently that means it won't be taken into account for
those two platforms. In the future, more .net versions will appear and the
current code should work. If we force the use of #*if (net45)* we could be
forced to include explicitly future labels for future platforms?

src/app/SharpRaven/Utilities/SystemUtil.cs
```diff
@@ -39,6 +40,19 @@ namespace SharpRaven.Utilities
...
     {
+#if net35
+        /// <summary>
+        /// Checks if a string is null or white space
+        /// </summary>
+        /// <param name="arg"></param>
+        /// <returns></returns>
+        public static bool IsNullOrWhiteSpace(string arg)
+        {
+            return string.IsNullOrEmpty(arg) || string.IsNullOrEmpty(arg.Trim());
```
asbjornu: It's fine to have and use this method in .NET 4.0 and 4.5 as well.

src/app/SharpRaven/packages.config
```diff
@@ -1,5 +1,4 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="GitVersionTask" version="3.4.1" targetFramework="net45" developmentDependency="true" />
-  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net40" />
```
asbjornu: Won't removing this break things?

src/packages/repositories.config
```diff
@@ -1,8 +0,0 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
```
asbjornu: Why is this file deleted?

src/app/SharpRaven/Utilities/CircularBuffer.cs
```diff
@@ -42,4 +45,105 @@ public bool IsEmpty()
             return queue.IsEmpty;
         }
     }
+
+    internal class ConcurrentQueue<T> : ICollection, IEnumerable<T>
```
asbjornu: If `ConcurrentQueue` is just an implementation detail of `CircularBuffer`, you can just put it in as a nested class. Otherwise, move it to its own file. Are you not missing an `#if (net35)` around the class declaration, by the way? We'd want to use `System.Collections.Concurrent.ConcurrentQueue` in .NET 4.0+ and not our custom implementation, no?

asbjornu:
@aldav82: Thanks for this! I've commented here and there and think that if everything I point out gets fixed, this can be merged. Great work! 😄  👍

src/app/SharpRaven/Data/Requester.Net45.cs
asbjornu: Excellent point. Keep it as is.

aldav82:
Thanks for your comments and for taking the time to check out these
changes. I see all your comments and they make perfect sense. However, I am
not sure how to follow from this point? I should continue making changes on
my fork and then create a new Pull Request? This is my first Pull Request
ever on Github, so I am not sure how to continue from this point.

asbjornu:
@aldav82: Just continue committing to your fork, `git push` the changes to GitHub and they will show up in this pull request automatically.

Preferably, you should have created a separate branch for your pull request (such as `net35` for instance), since you've now pull-requested your `develop` branch; inhibiting you to do any other work on the `develop` branch than whatever should be merged as a part of this pull request.

As long as you the commits you do until this PR is merged are related to the PR, we'll be fine, though.

src/app/SharpRaven/Data/SentryRequestFactory.cs
aldav82: The problem lies in the fact that the returned object will be an HttpContext, as defined in System.Web. When moving upwards, you are using dynamics, but that is one of the reasons the original project could not be compiled in .Net 3.5

asbjornu: Could the reference to `System.Web` be avoided if the `HttpContext` was treated as an `object` and then have its properties accessed through reflection instead? Too much work?

src/app/SharpRaven/RavenClient.cs
```diff
@@ -395,7 +406,12 @@ private string HandleException(Exception exception, Requester requester)
 
                 var response = webException.Response;
                 id = response.Headers["X-Sentry-ID"];
-                if (string.IsNullOrWhiteSpace(id))
+#if net35
+                var isNullOrWhiteSpace = Utilities.SystemUtil.IsNullOrWhiteSpace(id);
+#else
+                var isNullOrWhiteSpace = String.IsNullOrWhiteSpace(id);
+#endif
```
asbjornu: Can't you just use `Utilities.SystemUtil.IsNullOrWhiteSpace(id)` regardless here?

src/tests/SharpRaven.UnitTests/Data/JsonPacketFactoryTests.cs
```diff
@@ -115,7 +115,9 @@ public void Create_Project_EventIDIsValidGuid()
             var json = this.jsonPacketFactory.Create(project, (SentryMessage)null);
 
             Assert.That(json.EventID, Is.Not.Null.Or.Empty, "EventID");
+#if (!net35)
             Assert.That(Guid.Parse(json.EventID), Is.Not.Null);
+#endif
```
asbjornu: What's not .NET 3.5 compatible about this? Can we change something to make it compatible with all target frameworks?

src/app/SharpRaven/RavenClient.cs
aldav82: Yes, but I thought you would prefer the original official .net implementation if it was available.

src/tests/SharpRaven.UnitTests/SchemaHelper.cs
```diff
@@ -72,11 +72,18 @@ public static string GetSchemaPath()
             var directory = new DirectoryInfo(Environment.CurrentDirectory);
             FileInfo file = null;
 
+#if (net35)
+            while (directory != null && directory.Exists &&
+                   (file = directory.GetFiles("*.json")
+                       .FirstOrDefault(f => f.FullName.EndsWith("schema.json"))) == null)
+                            directory = directory.Parent;
+#else
```
asbjornu: Can't we just do it the .NET 3.5 way for all target frameworks?

src/app/SharpRaven/Data/SentryRequestFactory.cs
aldav82: Maybe. Could you elaborate as for why System.Web reference might be a problem? Using reflection might be feasible, but there is one method in particular
```csharp
public ISentryRequest Create()
        {
...
}
```

that is expecting the result of **HttpContext** property and it works with the properties of the object (_HttpContext.Request.Url_, _HttpContext.Request.HttpMethod_ and so on) so we would have to use reflection there too, and that implies quite more coding. I'am willing to try, but wouldn't it be better to use System.Web for the moment and then, in another PR, remove it? Again, if you could explain why not using System.Web from the start, I think I'll understand it better.

src/app/SharpRaven/RavenClient.cs
asbjornu: I prefer as few `#if` statements in the code as possible. :smile:

src/tests/SharpRaven.UnitTests/SchemaHelper.cs
aldav82: Same reason as the other comment, I thought you would prefer the original official .net implementation if it was available.

asbjornu: Nah, I'm not fuzzy about it. We've already disabled C# 6 for the entire solution, so I don't really care how "modern" the code is as long as it works.

aldav82:
Yes, .NET 3.5 will be a part of the Nuget package. Nuget supports .net 3.5

src/tests/SharpRaven.UnitTests/Data/JsonPacketFactoryTests.cs
aldav82: Guid.Parse is not part of .Net 3.5, I changed it to a custom method, though.

src/app/SharpRaven/RavenClient.cs
```diff
@@ -325,17 +325,10 @@ public string CaptureEvent(Exception e, Dictionary<string, string> tags)
         /// <returns>The <see cref="JsonPacket"/> which should be sent to Sentry.</returns>
         protected internal virtual JsonPacket PreparePacket(JsonPacket packet)
         {
-#if net35
             var isNullOrWhiteSpace = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Logger);
             var isNullOrWhiteSpaceRelease = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Release);
             var isNullOrWhiteSpaceEnvironment = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Environment);
             var isNullOrWhiteSpaceLogger = Utilities.SystemUtil.IsNullOrWhiteSpace(Logger);
```
asbjornu: Now that the `#if`s are gone, we don't need these variables anymore, do we?

src/tests/SharpRaven.UnitTests/Data/JsonPacketTests.cs
```diff
@@ -136,9 +134,7 @@ public void Constructor_ProjectAndException_EventIDIsValidGuid()
             var json = new JsonPacket(project, new Exception("Error"));
 
             Assert.That(json.EventID, Is.Not.Null.Or.Empty, "EventID");
-#if (!net35)
-            Assert.That(Guid.Parse(json.EventID), Is.Not.Null);
-#endif
+            Assert.That(TestHelper.Parse(json.EventID), Is.Not.Null);
```
asbjornu: 👍

src/SharpRaven.sln
```diff
@@ -45,32 +37,59 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{F44F1C47-D
...
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 3.5|x86.ActiveCfg = Debug 3.5|Any CPU
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.0|Any CPU.ActiveCfg = Debug 4.0|Any CPU
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.0|Any CPU.Build.0 = Debug 4.0|Any CPU
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.0|Mixed Platforms.ActiveCfg = Debug 4.0|Any CPU
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.0|Mixed Platforms.Build.0 = Debug 4.0|Any CPU
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.0|x86.ActiveCfg = Debug 4.0|Any CPU
-		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.5|Any CPU.ActiveCfg = Debug 4.5|Any CPU
-		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.5|Any CPU.Build.0 = Debug 4.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.5|Any CPU.ActiveCfg = Debug 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Debug 4.5|Any CPU.Build.0 = Debug 3.5|Any CPU
```
asbjornu: These two build configurations look weird. `Debug 4.5` and `Debug 3.5` in the same configuration?

src/SharpRaven.sln
```diff
@@ -81,96 +100,190 @@ Global
...
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|Any CPU.ActiveCfg = Release 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|Any CPU.Build.0 = Release 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|Mixed Platforms.ActiveCfg = Release 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|Mixed Platforms.Build.0 = Release 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|x86.ActiveCfg = Release 3.5|Any CPU
+		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}.Release|x86.Build.0 = Release 3.5|Any CPU
```
asbjornu: Here's a few more build configurations that are inconsistent. Why is the version number missing in the part before the pipe?

src/SharpRaven.sln
```diff
@@ -81,96 +100,190 @@ Global
...
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.5|x86.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|Any CPU.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: More inconsistencies.

src/SharpRaven.sln
```diff
@@ -81,96 +100,190 @@ Global
...
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 4.5|x86.Build.0 = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|Any CPU.Build.0 = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|x86.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: Even more.

src/SharpRaven.sln
```diff
@@ -81,96 +100,190 @@ Global
...
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release 4.5|x86.Build.0 = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|Any CPU.Build.0 = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|x86.ActiveCfg = Release 4.5|Any CPU
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}.Release|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: And more. 😄

src/SharpRaven.sln
```diff
@@ -81,96 +100,190 @@ Global
...
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|Any CPU.ActiveCfg = Release|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|Any CPU.Build.0 = Release|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|Mixed Platforms.ActiveCfg = Release|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|Mixed Platforms.Build.0 = Release|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|x86.ActiveCfg = Release|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release|x86.Build.0 = Release|Any CPU
```
asbjornu: And yet more. 😄

src/app/SharpRaven/RavenClient.cs
```diff
@@ -325,14 +325,18 @@ public string CaptureEvent(Exception e, Dictionary<string, string> tags)
         /// <returns>The <see cref="JsonPacket"/> which should be sent to Sentry.</returns>
         protected internal virtual JsonPacket PreparePacket(JsonPacket packet)
         {
-            packet.Logger = string.IsNullOrWhiteSpace(packet.Logger)
-                            || (packet.Logger == "root" && !string.IsNullOrWhiteSpace(Logger))
+            var isNullOrWhiteSpace = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Logger);
+            var isNullOrWhiteSpaceRelease = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Release);
+            var isNullOrWhiteSpaceEnvironment = Utilities.SystemUtil.IsNullOrWhiteSpace(packet.Environment);
+            var isNullOrWhiteSpaceLogger = Utilities.SystemUtil.IsNullOrWhiteSpace(Logger);
```
asbjornu: No point using variables here anymore, is it?

src/app/SharpRaven/SharpRaven.csproj
```diff
@@ -66,12 +91,21 @@
...
-    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL">
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 3.5' Or $(Configuration) == 'Debug 3.5'">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.6.0.8\lib\net35\Newtonsoft.Json.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.0' Or $(Configuration) == 'Debug 4.0'">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.6.0.8\lib\net40\Newtonsoft.Json.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.5' Or $(Configuration) == 'Debug 4.5'">
```
asbjornu: Hm. Three references in a row for `Newtonsoft.Json` seems a bit excessive, don't you think?

asbjornu: Ah, it's a different `HintPath` for each target platform. I see.

src/tests/SharpRaven.UnitTests/SharpRaven.UnitTests.csproj
```diff
@@ -58,11 +79,27 @@
...
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\NSubstitute.1.8.0.0\lib\net35\NSubstitute.dll</HintPath>
+    </Reference>
+    <Reference Include="NSubstitute, Version=1.8.0.0, Culture=neutral, PublicKeyToken=92dd2e9066daa5ca, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.0' Or $(Configuration) == 'Debug 4.0'">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\NSubstitute.1.8.0.0\lib\net40\NSubstitute.dll</HintPath>
+    </Reference>
+    <Reference Include="NSubstitute, Version=1.8.0.0, Culture=neutral, PublicKeyToken=92dd2e9066daa5ca, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.5' Or $(Configuration) == 'Debug 4.5'">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\NSubstitute.1.8.0.0\lib\net40\NSubstitute.dll</HintPath>
```
asbjornu: There is no .NET 4.5 version of NSubstitute?

src/tests/SharpRaven.UnitTests/SharpRaven.UnitTests.csproj
```diff
@@ -58,11 +79,27 @@
...
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.6.0.8\lib\net35\Newtonsoft.Json.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.0' Or $(Configuration) == 'Debug 4.0'">
+      <SpecificVersion>False</SpecificVersion>
+      <HintPath>..\..\packages\Newtonsoft.Json.6.0.8\lib\net40\Newtonsoft.Json.dll</HintPath>
+    </Reference>
+    <Reference Include="Newtonsoft.Json, Version=6.0.0.0, Culture=neutral, PublicKeyToken=30ad4fe6b2a6aeed, processorArchitecture=MSIL" Condition="$(Configuration) == 'Release 4.5' Or $(Configuration) == 'Debug 4.5'">
       <SpecificVersion>False</SpecificVersion>
       <HintPath>..\..\packages\Newtonsoft.Json.6.0.8\lib\net40\Newtonsoft.Json.dll</HintPath>
```
asbjornu: There's no .NET 4.5 version of Newtonsoft.Json?

src/SharpRaven.sln
```diff
@@ -45,32 +37,59 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "docs", "docs", "{F44F1C47-D
...
+		Release 3.5|x86 = Release 3.5|x86
 		Release 4.0|Any CPU = Release 4.0|Any CPU
 		Release 4.0|Mixed Platforms = Release 4.0|Mixed Platforms
 		Release 4.0|x86 = Release 4.0|x86
 		Release 4.5|Any CPU = Release 4.5|Any CPU
 		Release 4.5|Mixed Platforms = Release 4.5|Mixed Platforms
 		Release 4.5|x86 = Release 4.5|x86
+		Release|Any CPU = Release|Any CPU
+		Release|Mixed Platforms = Release|Mixed Platforms
+		Release|x86 = Release|x86
```
asbjornu: Which version of the .NET framework does the version-less build configurations build against? Should we have them?

src/app/SharpRaven/Data/SentryRequestFactory.cs
asbjornu: Good point. The reason for not using `System.Web` is that we don't want that to be a hard requirement for the library. It should only pick up the HTTP context if it's available, which is it's done in the convoluted way it is. There's a lot of situations where you don't want or can't have a reference to `System.Web`, so that's why we should avoid it. Having it in .NET 3.5 might be a fine compromise, though.

asbjornu:
@aldav82 Ok, good. I have a few more comments that I would like you to address, but this is starting to look merge-able. 😄

src/SharpRaven.sln
aldav82: We shouldn't. I fixed the solution file. I think it's ok now. Let me know

src/tests/SharpRaven.UnitTests/SharpRaven.UnitTests.csproj
aldav82: Yes, there is. Fixed in cedaf95ec3b3e70aa98019389ef5f26d51d7411c.

src/tests/SharpRaven.UnitTests/SharpRaven.UnitTests.csproj
aldav82: Fixed in 	77cc6c3

src/app/SharpRaven/RavenClient.cs
aldav82: Correct. Fixed c841293

src/SharpRaven.sln
```diff
@@ -81,96 +97,160 @@ Global
...
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Debug 4.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Debug 4.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Debug 4.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Debug 4.5|x86.Build.0 = Debug 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|Any CPU.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 3.5|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: `3.5` and `4.5` in the same build configuration? :smiley:

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
     <TargetFrameworkProfile />
     <NuGetPackageImportStamp>ec5b9b6c</NuGetPackageImportStamp>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug 3.5|AnyCPU'">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\net40\</OutputPath>
```
asbjornu: Should `Debug 3.5` output to the `net40` folder?

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
...
+    <DefineConstants>TRACE;DEBUG;net40</DefineConstants>
```
asbjornu: Should `Debug 3.5` define the `net40` constant?

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
...
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>bin\Debug\net40\SharpRaven.Nancy.xml</DocumentationFile>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <Prefer32Bit>false</Prefer32Bit>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
```
asbjornu: Shouldn't this be `v3.5`?

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release 3.5|AnyCPU'">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\net40\</OutputPath>
+    <DefineConstants>TRACE;net40</DefineConstants>
```
asbjornu: Same as above.

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
...
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>bin\Release\net40\SharpRaven.Nancy.xml</DocumentationFile>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <Prefer32Bit>false</Prefer32Bit>
+    <TargetFrameworkVersion>v4.0</TargetFrameworkVersion>
```
asbjornu: Same as above.

src/app/SharpRaven.Nancy/SharpRaven.Nancy.csproj
```diff
@@ -15,6 +15,31 @@
...
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release 3.5|AnyCPU'">
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\net40\</OutputPath>
+    <DefineConstants>TRACE;net40</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>bin\Release\net40\SharpRaven.Nancy.xml</DocumentationFile>
```
asbjornu: Same as above.

src/app/SharpRaven/IRavenClient.cs
```diff
@@ -215,7 +217,7 @@ public interface IRavenClient
 
 #endif
 
-        #region Deprecated Methods
+#region Deprecated Methods
```
asbjornu: I think I like it better when all of the preprocessor directives (`#if`, `#region`, etc.) are aligned with the code they surround and not left-aligned. Can you please revert this and properly align all the `#if` processor directives you've added?

src/app/SharpRaven/RavenClient.cs
```diff
@@ -325,14 +325,14 @@ public string CaptureEvent(Exception e, Dictionary<string, string> tags)
...
+            packet.Logger = SystemUtil.IsNullOrWhiteSpace(packet.Logger)
+                            || (packet.Logger == "root" && !SystemUtil.IsNullOrWhiteSpace(Logger))
                 ? Logger
                 : packet.Logger;
             packet.User = packet.User ?? this.sentryUserFactory.Create();
             packet.Request = packet.Request ?? this.sentryRequestFactory.Create();
-            packet.Release = string.IsNullOrWhiteSpace(packet.Release) ? Release : packet.Release;
-            packet.Environment = string.IsNullOrWhiteSpace(packet.Environment) ? Environment : packet.Environment;
+            packet.Release = SystemUtil.IsNullOrWhiteSpace(packet.Release) ? Release : packet.Release;
+            packet.Environment = SystemUtil.IsNullOrWhiteSpace(packet.Environment) ? Environment : packet.Environment;
```
asbjornu: :+1:

src/app/SharpRaven/RavenClient.cs
```diff
@@ -395,7 +395,8 @@ private string HandleException(Exception exception, Requester requester)
 
                 var response = webException.Response;
                 id = response.Headers["X-Sentry-ID"];
-                if (string.IsNullOrWhiteSpace(id))
+                var isNullOrWhiteSpace = Utilities.SystemUtil.IsNullOrWhiteSpace(id);
+                if (isNullOrWhiteSpace)
```
asbjornu: No need for a variable here, is it?

src/app/SharpRaven/SharpRaven.csproj
```diff
@@ -15,6 +15,31 @@
...
   </PropertyGroup>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug 3.5|AnyCPU'">
+    <DebugSymbols>true</DebugSymbols>
+    <DebugType>full</DebugType>
+    <Optimize>false</Optimize>
+    <OutputPath>bin\Debug\net35\</OutputPath>
+    <DefineConstants>TRACE;DEBUG;net35</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+    <DocumentationFile>bin\Debug\net35\SharpRaven.xml</DocumentationFile>
```
asbjornu: Looks like you got the version number right here. :smiley: :+1:

src/app/SharpRaven/Utilities/CircularBuffer.cs
```diff
@@ -1,11 +1,121 @@
...
+#endif
+using System.Collections;
 using System.Collections.Generic;
 using System.Linq;
 
 namespace SharpRaven.Utilities {
     public class CircularBuffer<T>
     {
+#if net35
+        private class ConcurrentQueue<K> : ICollection, IEnumerable<K>
```
asbjornu: Why `K` as the generic type parameter and not `T`?

src/tests/SharpRaven.Nancy.UnitTests/SharpRaven.Nancy.UnitTests.csproj
```diff
@@ -15,7 +15,7 @@
     <TargetFrameworkProfile />
     <NuGetPackageImportStamp>a120c3a0</NuGetPackageImportStamp>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release 4.0|AnyCPU'">
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release 3.5|AnyCPU'">
```
asbjornu: Should the version number change from `4.0` to `3.5` here?

src/tests/SharpRaven.Nancy.UnitTests/SharpRaven.Nancy.UnitTests.csproj
```diff
@@ -25,9 +25,19 @@
     <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
     <Prefer32Bit>true</Prefer32Bit>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release 4.5|AnyCPU'">
-    <OutputPath>bin\Release\net45\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug 3.5|AnyCPU'">
+    <DebugSymbols>true</DebugSymbols>
+    <OutputPath>bin\Debug\net40\</OutputPath>
+    <DefineConstants>TRACE;DEBUG;net40</DefineConstants>
```
asbjornu: The build configuration says `3.5`, but the constants and paths say `net40`.

src/tests/SharpRaven.UnitTests/Data/HttpRequestBodyConverterTests.cs
```diff
@@ -43,14 +45,18 @@
...
     public class HttpRequestBodyConverterTests
     {
         [Test]
         public void Convert_Form_ReturnsForm()
         {
+
             dynamic httpContext = new ExpandoObject();
             httpContext.Request = new ExpandoObject();
+
+
```
asbjornu: What's all this whitespace doing there? :smile:

src/tests/SharpRaven.UnitTests/packages.config
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
   <package id="GitVersionTask" version="3.4.1" targetFramework="net45" developmentDependency="true" />
-  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net40" />
-  <package id="NSubstitute" version="1.8.0.0" targetFramework="net40" />
+  <package id="Newtonsoft.Json" version="6.0.8" targetFramework="net40"/>
+  <package id="NSubstitute" version="1.8.0.0" targetFramework="net40"/>
```
asbjornu: Should the `targetFramework` be consistent for all packages, perhaps?

src/app/SharpRaven/Utilities/CircularBuffer.cs
aldav82: Fixed 948341e132cc7284a19eaa7fb9c06e188fdbce75

src/app/SharpRaven/RavenClient.cs
aldav82: Fixed 948341e132cc7284a19eaa7fb9c06e188fdbce75

src/app/SharpRaven/IRavenClient.cs
aldav82: Fixed c552f6c

src/tests/SharpRaven.UnitTests/packages.config
aldav82: Fixed 6bf6594

src/SharpRaven.sln
```diff
@@ -81,96 +97,148 @@ Global
...
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Any CPU.ActiveCfg = Debug 4.0|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Any CPU.Build.0 = Debug 4.0|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|x86.Build.0 = Debug 4.5|Any CPU
```
asbjornu: `3.5` and `4.5` mixed again. :smile:

src/SharpRaven.sln
```diff
@@ -81,96 +97,148 @@ Global
...
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|x86.Build.0 = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Any CPU.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: Same.

src/app/SharpRaven/RavenClient.cs
```diff
@@ -325,14 +325,19 @@ public string CaptureEvent(Exception e, Dictionary<string, string> tags)
...
                 : packet.Logger;
             packet.User = packet.User ?? this.sentryUserFactory.Create();
-            packet.Request = packet.Request ?? this.sentryRequestFactory.Create();
-            packet.Release = string.IsNullOrWhiteSpace(packet.Release) ? Release : packet.Release;
-            packet.Environment = string.IsNullOrWhiteSpace(packet.Environment) ? Environment : packet.Environment;
+            try {
+                packet.Request = packet.Request ?? this.sentryRequestFactory.Create();
+            }
+            catch {
+                packet.Request = this.sentryRequestFactory.Create();
```
asbjornu: If the above `sentryRequestFactory.Create()` failed, won't this fail as well? I don't understand this `try/catch`.

src/tests/SharpRaven.UnitTests/Data/HttpRequestBodyConverterTests.cs
```diff
@@ -43,12 +45,14 @@
 
 namespace SharpRaven.UnitTests.Data
 {
+    #if (!net35)
```
asbjornu: What is it about these tests that don't work on .NET 3.5?

aldav82: The ExpandoObject class

src/app/SharpRaven/RavenClient.cs
aldav82: Yes, it's true. I assigned null in 187a10e . The goal of the try/catch is to address a very specific exception I ran into, _trying to access Request in Application_Start in IIS7 Integrated Mode_ (http://mvolo.com/iis7-integrated-mode-request-is-not-available-in-this-context-exception-in-applicationstart/)

src/SharpRaven.sln
aldav82: Visual Studio keeps messing with the solution file and mixing up the configurations. I think now it uploaded correctly. I keep fixing them and Visual Studio keeps changing them before the commit.

src/SharpRaven.sln
```diff
@@ -81,96 +97,136 @@ Global
...
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Any CPU.ActiveCfg = Release 4.0|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Any CPU.Build.0 = Release 4.0|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Mixed Platforms.ActiveCfg = Release 4.0|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Mixed Platforms.Build.0 = Release 4.0|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|x86.ActiveCfg = Release 4.0|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
-		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
```
asbjornu: Major shuffle! :smile: 

shuffle

src/tests/SharpRaven.UnitTests/Data/HttpRequestBodyConverterTests.cs
asbjornu: Ah, of course.

src/SharpRaven.sln
aldav82: Sorry, are we looking at the same changeset? The configurations I removed where the ones for .net 3.5 in Nancy project, because Nancy, as it is, doesn't support .Net 3.5. I don't quite get in which changeset was the change you mention.

src/SharpRaven.sln
```diff
@@ -81,96 +97,136 @@ Global
...
 	GlobalSection(NestedProjects) = preSolution
 		{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1} = {88CB5CD6-ACFA-441E-9FBB-E1E06F181B89}
-		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31} = {D2B3990A-A048-49F3-A697-A1DAEF82CE7F}
-		{156621FC-2C48-4CDF-A368-9347BABE9089} = {D2B3990A-A048-49F3-A697-A1DAEF82CE7F}
+		{F44F1C47-DCF1-472A-A2D1-7F132AA54339} = {BF6154D1-9BC4-4966-9506-BC33E81F317C}
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3} = {88CB5CD6-ACFA-441E-9FBB-E1E06F181B89}
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C} = {D2B3990A-A048-49F3-A697-A1DAEF82CE7F}
-		{F44F1C47-DCF1-472A-A2D1-7F132AA54339} = {BF6154D1-9BC4-4966-9506-BC33E81F317C}
+		{E1DBEBBF-9448-4D99-B378-2B8CF1629F31} = {D2B3990A-A048-49F3-A697-A1DAEF82CE7F}
+		{156621FC-2C48-4CDF-A368-9347BABE9089} = {D2B3990A-A048-49F3-A697-A1DAEF82CE7F}
```
aldav82: I am guessing Visual Studio? I am editing the solution file in Notepad++ while I have it opened in Visual Studio.

src/SharpRaven.sln
asbjornu: I'm not looking at a particular commit, just the entire diff of the PR as presented by GitHub and there's two project identifiers affected here: `{E1DBEBBF-9448-4D99-B378-2B8CF1629F31}` and `{156621FC-2C48-4CDF-A368-9347BABE9089}`. It seems like they are just moved further down into the `.sln` file. I'll reference that part in another comment.

src/SharpRaven.sln
```diff
@@ -81,96 +97,136 @@ Global
...
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Mixed Platforms.ActiveCfg = Release 4.0|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|Mixed Platforms.Build.0 = Release 4.0|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|x86.ActiveCfg = Release 4.0|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.0|x86.Build.0 = Release 4.0|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Any CPU.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: Here's the stuff that seems to have moved from [what is referenced in this comment](https://github.com/getsentry/raven-csharp/pull/156/files#r79582300).

src/SharpRaven.sln
asbjornu: Can you please revert changes that are basically just moving lines up/down in the `.sln` file?

src/SharpRaven.sln
aldav82: I see. I made the changes in this commit 60f9f22

asbjornu:
@aldav82: Excellent, thanks for this! Pushing out a new package on NuGet as soon as possible to see how it works.

asbjornu:
[2.1.1-unstable0107](https://www.nuget.org/packages/SharpRaven.Nancy/2.1.1-unstable0107) published.

aldav82:
Thanks a lot! I'll start testing it right away!

PR #159, opened 2016-09-19
Feature public sentryevent.tags

samukce:
Complementing the PR #149. Some tests case to ensure public assign from SentryEvent.Tags

---
This change is [Reviewable](https://reviewable.io/reviews/getsentry/raven-csharp/159)

### DIFF
--- a/src/tests/SharpRaven.UnitTests/Data/SentryEventTests.cs
+++ b/src/tests/SharpRaven.UnitTests/Data/SentryEventTests.cs
@@ -1,0 +1,77 @@
+﻿#region License
+
+// Copyright (c) 2014 The Sentry Team and individual contributors.
+// All rights reserved.
+// 
+// Redistribution and use in source and binary forms, with or without modification, are permitted
+// provided that the following conditions are met:
+// 
+//     1. Redistributions of source code must retain the above copyright notice, this list of
+//        conditions and the following disclaimer.
+// 
+//     2. Redistributions in binary form must reproduce the above copyright notice, this list of
+//        conditions and the following disclaimer in the documentation and/or other materials
+//        provided with the distribution.
+// 
+//     3. Neither the name of the Sentry nor the names of its contributors may be used to
+//        endorse or promote products derived from this software without specific prior written
+//        permission.
+// 
+// THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR
+// IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND
+// FITNESS FOR A PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR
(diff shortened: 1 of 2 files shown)


samukce:
@asbjornu done. o/

asbjornu:
@samukce Excellent! But can you perhaps remove the three merge commits from this PR? This is rather easy to do by resetting this branch to the latest `develop` and then doing `cherry-pick` to pick the interesting commit from here. Like so:

```shell
git fetch getsentry
git checkout feature/publicTagsSentryEvent
git reset --hard getsentry/develop
git cherry-pick 8d868bf5417ee8f5824faf1102d05306f4708278
git push --force
```

samukce:
@asbjornu thanks for the feedback. Done.

src/tests/SharpRaven.UnitTests/Data/SentryEventTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            Assert.That(sentryEvent.Tags, Is.Not.Null);
+        }
+
+        [Test]
+        public void ShouldTheSameDictionaryFromConstructor()
+        {
+            var sentryEvent = new SentryEvent(new Exception("foo"));
+            var beforeDictionary = sentryEvent.Tags;
+
+            Assert.That(beforeDictionary, Is.SameAs(sentryEvent.Tags));
```
asbjornu: I don't quite see the value of this assertion.

samukce: The get/set execute some procedures how instance new dictionary. Some refactor could inject a bug, for instance, a new dictionary, what could losing data.

asbjornu: Good point!

src/tests/SharpRaven.UnitTests/Data/SentryEventTests.cs
```diff
@@ -0,0 +1,77 @@
...
+        {
+            Assert.That(new SentryEvent(new Exception("foo")).Tags, Is.Not.Null);
+        }
+
+        [Test]
+        public void ShouldNotNullDictionaryWhenAssignNullForTags()
+        {
+            var sentryEvent = new SentryEvent(new Exception("foo")) { Tags = null};
+
+            Assert.That(sentryEvent.Tags, Is.Not.Null);
```
genne: Regarding this. The test case is great, but the "feature" feels a bit weird. I would really prefer if the setter of `Tags` threw an `ArgumentNullException` or similar when trying to set it to `null`. Setting something to `null` and then making sure it wasn't set to `null` doesn't really make sense?

asbjornu: @genne: Good point. Let's do `ArgumentNullException`, then. PR incoming? :smile:

samukce: I think that this api should work similar to the log4net, i mean when any is wrong with the API, this not interfere of the function of the software.

asbjornu: @samukce While I understand that design principle from a logging framework, there's already quite a few `throw` statements in there.

samukce: @asbjornu  but do you agree that some simple exceptions could be avoided?

asbjornu: Yes, but in this case, I'm basically ¯\_(ツ)_/¯.

PR #164, opened 2016-10-04
Demangle Function Name from Async Calls

### LINKED ISSUE #147: Ideas on demangling the function names of async tasks?
Hi there,

I'm currently trying to clean up how our C# stack-traces look in Sentry. You merged my PR last week to add the 'in-app' flag which has made a _huge_ difference. Now we only see our code by default in stack traces.

The next piece I want to try to tackle is the names of async functions, but I thought it best to open an issue before writing any code.

Our stack-traces (from the dotnet runtime) have items that look like this: (and they appear the same in Sentry:)

`Flipdish.Web.Controllers.PollApiController+<RequestNewOrders>d__4 in MoveNext at line 37:13`

Which comes from this code:

```C#
public async Task<ActionResult> RequestNewOrders()
{
  var requestTime = System.Web.HttpContext.Current.Timestamp;
  var userId = User.Identity.GetUserId<int>();
  // The following is line 37, which is 'MoveNext' I guess:
  IEnumerable<PreviousOrderDm> result = await _posApiService.RequestNewOrders(requestTime, userId); 
// ...
}

```

At first glance, it looks like the C# compiler is creating a little function with a random name in place of the `async`/`await` syntactic sugar, which results in the method name `PollApiController+<RequestNewOrders>d__4 in MoveNext`.

I wonder is there a way to get a nicer looking name than that?

In the past I have worked on Clojure projects and run into the same problem. The Clojure compiler creates JVM function with similar names; Clojure functions can have characters in the function name that the JVM does not allow, such as `?` and `!`. The Clojure compiler will replace these characters with `_QMARK_` and `_BANG`. A function named `list?` in Clojure will be named `list_QMARK_` when compiled for the JVM. Clojure exception loggers will typically have a demangling step which will reverse these replacements in the function names, so that the function name reported to Sentry, for example, would be `list?`. The happens in the exception logging client before the JSON is sent to the logger.

Here is such a look-up table.

https://github.com/mmcgrana/clj-stacktrace/blob/master/src/clj_stacktrace/core.clj#L22-L34

I wonder if it would be possible to do the same in raven-csharp?

We might be able to change:

`Flipdish.Web.Controllers.PollApiController+<RequestNewOrders>d__4 in MoveNext at line 37:13`

to:

`Flipdish.Web.Controllers.PollApiController.RequestNewOrders in async task at line 37:13`

What do you think? Good idea/crazy idea?

marcomorain:
When the Microsoft cs.exe compiler compiles some modern C# features,
such as async/await calls, it can create synthetic function names that
do not match the function names in the original source code. Here we
reverse some of these transformations, so that the function and module
names that appears in the Sentry UI will match the function and module
names in the original source-code.

Fixes #147

### DIFF
--- a/src/app/SharpRaven/Data/ExceptionFrame.cs
+++ b/src/app/SharpRaven/Data/ExceptionFrame.cs
@@ -33,6 +33,7 @@ using System.Diagnostics;
 using System.Text;
 
 using Newtonsoft.Json;
+using System.Text.RegularExpressions;
 
 namespace SharpRaven.Data
 {
@@ -81,6 +82,7 @@ namespace SharpRaven.Data
             LineNumber = lineNo;
             ColumnNumber = frame.GetFileColumnNumber();
             InApp = !IsSystemModuleName(Module);
+            DemangleFunctionName();
         }
 
 
@@ -118,7 +120,7 @@ namespace SharpRaven.Data
         /// The function.
         /// </value>
         [JsonProperty(PropertyName = "function")]
-        public string Function { get; set; }
+        public string Function { get; private set; }
 
         /// <summary>
         /// Signifies whether this frame is related to the execution of the relevant code in this
@@ -148,7 +150,7 @@ namespace SharpRaven.Data
         /// The module.
         /// </value>
         [JsonProperty(PropertyName = "module")]
-        public string Module { get; set; }
+        public string Module { get; private set; }
 
         /// <summary>
         /// Gets or sets the post context.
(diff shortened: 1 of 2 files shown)


src/app/SharpRaven/Data/ExceptionFrame.cs
```diff
@@ -227,8 +229,42 @@ public override string ToString()
...
+        /// </summary>
+        /// <para>
+        /// When the Microsoft cs.exe compiler compiles some modern C# features,
+        /// such as async/await calls, it can create synthetic function names that
+        /// do not match the function names in the original source code. Here we
+        /// reverse some of these transformations, so that the function and module
+        /// names that appears in the Sentry UI will match the function and module
+        /// names in the original source-code.
+        /// </para>
+        private void demangleFunctionName()
```
asbjornu: Uppercase first letter in method names, please. 😃

src/tests/SharpRaven.UnitTests/Data/ExceptionFrameTests.cs
```diff
@@ -77,19 +78,49 @@ public void Constructor_InAppFrames_Identified()
...
                 var frames = new StackTrace(exception, true).GetFrames();
                 var sentryFrames = frames.Select(f => new ExceptionFrame(f)).ToList();
 
                 // The first frame is InApp - (this function).
-                Assert.True(sentryFrames.First().InApp); 
+                Assert.True(sentryFrames.First().InApp);
 
                 // The last frame is InApp (our lambda that divides by zero).
-                Assert.True(sentryFrames.Last().InApp); 
+                Assert.True(sentryFrames.Last().InApp);
```
asbjornu: Try to avoid touching code that isn't a part of the problem this PR is fixing, please. 😃

marcomorain:
@asbjornu ok - should be ready to rock now 🚀

asbjornu:
@marcomorain: Awesome, thanks!

marcomorain:
Thanks! Im looking forward to deploying this.





On Wed, 5 Oct 2016, at 08:33, Asbjørn Ulsberg wrote:
> Merged #164[1].
> — You are receiving this because you were mentioned. Reply to this
> email directly, view it on GitHub[2], or mute the thread[3].
>


Links:

  1. https://github.com/getsentry/raven-csharp/pull/164
  2. https://github.com/getsentry/raven-csharp/pull/164#event-812855780

PR #165, opened 2016-10-06
Improve exception frame tests

marcomorain:
Add second test case for function demangling.
Improve formatting of constructor tests.


marcomorain:
This applies some of the clean-up that we didn't commit in my last PR.

PR #166, opened 2016-10-27
Demangle Anonymous Function Names

### LINKED ISSUE #147: Ideas on demangling the function names of async tasks?

marcomorain:
Demangle compiler generated function names for anonymous functions and
replace with something more readable.

Change:

```
<BeginInvokeAsynchronousActionMethod>b__36
```

to:

```
BeginInvokeAsynchronousActionMethod { λ }
```

Fixes #147

### DIFF
--- a/src/app/SharpRaven/Data/ExceptionFrame.cs
+++ b/src/app/SharpRaven/Data/ExceptionFrame.cs
@@ -82,7 +82,8 @@ namespace SharpRaven.Data
             LineNumber = lineNo;
             ColumnNumber = frame.GetFileColumnNumber();
             InApp = !IsSystemModuleName(Module);
-            DemangleFunctionName();
+            DemangleAsyncFunctionName();
+            DemangleAnonymousFunction();
         }
 
 
@@ -244,7 +245,7 @@ namespace SharpRaven.Data
         /// names that appears in the Sentry UI will match the function and module
         /// names in the original source-code.
         /// </para>
-        private void DemangleFunctionName()
+        private void DemangleAsyncFunctionName()
         {
             if (Module == null || Function != "MoveNext")
             {
@@ -264,6 +265,31 @@ namespace SharpRaven.Data
             {
                 Module = match.Groups[1].Value;
                 Function = match.Groups[2].Value;
+            }
+        }
+
+        /// <summary>
+        /// Clean up function names for anonymous lambda calls.
+        /// </summary>
+        private void DemangleAnonymousFunction()
+        {
+            if (Function == null)
+            {
+                return;
+            }
+
+            // Search for the function name in angle brackets followed by b__<digits/letters>.
+            //
+            // Change:
+            //   <BeginInvokeAsynchronousActionMethod>b__36
+            // to:
+            //   BeginInvokeAsynchronousActionMethod { <lambda> }
+
+            var mangled = @"^<(\w*)>b__\w+$";
+            var match = Regex.Match(Function, mangled);
+            if (match.Success && match.Groups.Count == 2)
+            {
+                Function = match.Groups[1].Value + " { <lambda> }";
             }
         }
     }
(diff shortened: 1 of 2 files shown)


asbjornu:
@marcomorain Thanks for this! But as you mention in #147, it would be awesome with more intelligence in the demangling of the anonymous method, which makes me think `λ` might look a bit esoteric and not very compatible with future improvements. I also asked a colleague and while he appreciated it, he thought it looked a bit too "smart". If we're able to figure out arguments, return value and such things, I think `BeginInvokeAction { <lambda> }` is a good place to start and then expand to `() ⇒ …` once we get `(x, y) => z` to work.

I'm also wondering about the `BeginInvokeAction`; is that a constant method name that will always exist on `async` methods? If so, we should probably replace that with the `async` keyword as well, no?

marcomorain:
I'll change this to use  `<lambda>` rather than the greek symbol - I don't have a strong preference myself, and I think you are right - the greek symbol is us trying to be a bit too clever!

> I'm also wondering about the BeginInvokeAction; is that a constant method name that will always exist on async methods? If so, we should probably replace that with the async keyword as well, no?

I don't have access to the ASP.net MVC source, but I'm very confident that `BeginInvokeAction` is the actual function name for invoking controller actions, rather than a compiler generated name.

marcomorain:
@asbjornu updated to use `<lambda>` name

asbjornu:
@marcomorain: Thanks!

marcomorain:
👍

asbjornu:
[Version 2.1.1-unstable0113](http://www.nuget.org/packages/SharpRaven/2.1.1-unstable0113) released with this folded in. If I don't hear anything negative about this release for the next couple of days, I'll try to push out a new stable by the end of the week.

marcomorain:
I've deployed this to production, I'll let you know if we have any problems. If you don't hear from me it's going well!

PR #168, opened 2016-12-14
Just Another .NET Core port

phenixdotnet:
Hi.

I'm aware of the great work do by @wakawaka54 and @reflectiondm to port raven-sharp to .NET core but I started my own port because:
- I worked on several port to .NET core and I prefer to keep only one project which build all version rather than to keep separate projects.
- I added integration of raven into AspNet Core filter exception to enable automatic errors reporting
- I added 2 samples projects, one with .NET Core app and one with AspNetCore project
- I added support of cake build script
- It's fun ^^ 

It's not perfect (I didn't work on unit tests actually) but its work except the if compression is enabled.

I started the merge request to gets your feedback on this, any comments/remarks are welcome :)

Regards


asbjornu:
@phenixdotnet: Thank you so much for the incredible work you've put into this! As you seem to have gotten much closer to implementing Cake (as I started in #104), I would love to have that as a separate pull request unrelated to .NET Core. Would it be possible for you to do that, so we can merge it and have this pull request focus just on .NET Core and building on the (hopefully already working) Cake implementation?

phenixdotnet:
@asbjornu It's possible but I don't see benefit to use Cake if you don't go cross platform. Do you have any scenario in mind ?
Actually I reworked the travis build script to work with Cake, only unit tests failed but its not related to using Cake

asbjornu:
@phenixdotnet: The build is and has been working on Linux (Travis) for quite a few months, so with Mono, the current version is xplat already.

phenixdotnet:
@asbjornu Of course, using MSBuild / XBuild will required to rework the Cake script, not a big challenge but you have to maintains 2 paths : One for windows build, one for unix build. 
.NET core build is simpler because you have only one set of commands.

So if you want it now I can have a look at it otherwise I suggest to wait for .NET core merge and have one build script to rule them all ;)

asbjornu:
@phenixdotnet: Can we use `dotnet.exe` to build SharpRaven for version 3.5 of .NET? If so, there's not much value in keeping the existing build pipeline, but if we need forks for legacy .NET and .NET Core, I think it would be valuable to have a Cake based build pipeline regardless.

phenixdotnet:
@asbjornu The .NET Core branch target the following framework version:

- netstandard16
- net46
- net45
- net40
- net35

The unit tests target only net46 and netstandard16 due to incompatible dependencies.

asbjornu:
@phenixdotnet: I see. So we could, in theory, migrate the whole build over to .NET Core without sacrificing backwards compatibility? What would be needed to get the unit tests executed for every desired target framework?

phenixdotnet:
@asbjornu Yes for the main lib (it's already the case in netcore branch, the build produce a nupkg with a version for each frameworks).

ravensharp_nupkg


The unit tests is compatible only with netstandard1.6 and .net 46 because of the dotnet-test-nunit and Microsoft.DotNet.InternalAbstractions so you can't run unit tests over .NET 3.5 version.

asbjornu:
@phenixdotnet: Okay, cool. Can't we use a different NUnit test runner depending on which .NET framework we are testing for? What's `Microsoft.DotNet.InternalAbstractions` good for?

phenixdotnet:
@asbjornu unfortunately no, the Nunit test runner is linked to dotnet sdk so you should use a one compatible with the sdk version. 

And ``Microsoft.DotNet.InternalAbstractions`` is a required dependencies to run unit tests. It has been added in .NET Core 1.1.0, may be it will be removed on the next release as it should be needed to reference it from ours projects.

asbjornu:
I hope #197 gives us the .NET Core / Standard support we need, so I'm closing this. Please `rebase` on `HEAD` of `develop` and reopen if there's something you feel is missing or has room for improvement. Thanks!

PR #172, opened 2017-01-13
Added user injection to SentryEvent. Also added offline caching of jsonpackets

rolandh:
I updated the constructor of the SentryEvent to accept and optional User field. If it is populated the user is injected into the jsonPacket. If it is null or empty the original code is executed and the username is populated from the currently logged in user.

I also added offline caching of packets to a .json file if no internet connection is available. Next time the client is initialised the packets are sent to the server.

The message field is updated to include the UTC time of which the packet send failed previously. This is done every time the client is loaded.

The concept could be easily be expanded to take a serialiser class that allows the missed packets to be stored differently. It currently meets my own requirements, though you may wish to refactor/improve it.

edit: Automated compile suggests compile error on the unit tests, I did not edit these classes so not sure what the deal is there.

### DIFF
--- a/src/SharpRaven.sln
+++ b/src/SharpRaven.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 2013
-VisualStudioVersion = 12.0.40629.0
+# Visual Studio 14
+VisualStudioVersion = 14.0.25420.1
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "SharpRaven", "app\SharpRaven\SharpRaven.csproj", "{CC80A2E1-AE39-44DE-8DA3-4EEF42F90FB1}"
 EndProject
@@ -161,6 +161,12 @@ Global
 		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
 		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
 		{156621FC-2C48-4CDF-A368-9347BABE9089}.Release 4.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Any CPU.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Any CPU.Build.0 = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 3.5|x86.Build.0 = Debug 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.0|Any CPU.ActiveCfg = Debug 4.0|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.0|Any CPU.Build.0 = Debug 4.0|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.0|Mixed Platforms.ActiveCfg = Debug 4.0|Any CPU
@@ -171,6 +177,12 @@ Global
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Debug 4.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Any CPU.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 3.5|x86.Build.0 = Release 4.5|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.0|Any CPU.ActiveCfg = Release 4.0|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.0|Any CPU.Build.0 = Release 4.0|Any CPU
 		{ABE22746-6EEB-4970-A608-C02BC3B8BDA3}.Release 4.0|Mixed Platforms.ActiveCfg = Release 4.0|Any CPU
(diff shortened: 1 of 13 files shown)


src/SharpRaven.sln
```diff
@@ -191,6 +209,12 @@ Global
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Debug 4.5|Mixed Platforms.ActiveCfg = Debug 4.5|Any CPU
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Debug 4.5|Mixed Platforms.Build.0 = Debug 4.5|Any CPU
 		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Debug 4.5|x86.ActiveCfg = Debug 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|Any CPU.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|Any CPU.Build.0 = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|Mixed Platforms.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|Mixed Platforms.Build.0 = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|x86.ActiveCfg = Release 4.5|Any CPU
+		{7FE52A2B-430C-4C6F-BEA9-0855AF973D0C}.Release 3.5|x86.Build.0 = Release 4.5|Any CPU
```
asbjornu: Are all of these solution changes necessary?

src/app/SharpRaven/Data/Breadcrumb.cs
```diff
@@ -9,6 +9,7 @@ namespace SharpRaven.Data {
     /// <summary>
     /// Breadcrumb trail.
     /// </summary>
+    [Serializable]
```
asbjornu: Why does `Breadcrumb` have to be `Serializable`?

src/app/SharpRaven/Data/JsonPacket.cs
```diff
@@ -396,5 +397,10 @@ private void Initialize(Exception exception)
                 }
             }
         }
+
+        public override int GetHashCode()
+        {
+            return this.ToString().GetHashCode()*17;
```
asbjornu: Why is this override necessary, and if it is, can't it just return `ToString().GethashCode()` directly? Why multiply by 17?

src/app/SharpRaven/Data/JsonPacketFactory.cs
```diff
@@ -130,9 +135,11 @@ public JsonPacket Create(string project, SentryEvent @event)
         {
             var json = new JsonPacket(project, @event)
             {
-                Breadcrumbs = @event.Breadcrumbs
+                Breadcrumbs = @event.Breadcrumbs,
```
asbjornu: Why did you add this comma? 😄

src/app/SharpRaven/Data/SentryEvent.cs
```diff
@@ -36,29 +36,34 @@ namespace SharpRaven.Data
     /// <summary>
     /// Represents an event being captured by <see cref="IRavenClient.Capture(SentryEvent)"/>.
     /// </summary>
+    [Serializable]
```
asbjornu: Why does `SentryEvent` need to be `Serializable`?

src/app/SharpRaven/RavenClient.cs
```diff
@@ -363,6 +393,19 @@ protected virtual string Send(JsonPacket packet)
...
+                if (!cachePacketsWhenOffline) return HandleException(exception, requester);
+                try
+                {
+                    if (!string.IsNullOrEmpty(packet.Message)) packet.Message += "\n";
+                    packet.Message +=
+                        $"Attempted to send on {DateTime.UtcNow.ToShortDateString()} {DateTime.UtcNow.ToLongTimeString()} UTC however packet failed to send and was cached locally.";
+                    var bytes = packet.SerialiseBytes();
+                    var fileName =
+                        $"{DateTime.UtcNow.ToShortDateString().Replace("/", "-")} {DateTime.UtcNow.ToLongTimeString().Replace(":", "-")} - {GetHashCode()}.json";
+                    var path = Path.Combine(System.Environment.CurrentDirectory, fileName);
```
asbjornu: I think the folder should be based around the DSN, a temporary folder, or something similar. In a web context, `CurrentDirectory` might not be writable.

src/app/SharpRaven/Utilities/Serializer.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace SharpRaven.Utilities
+{
+    public static class Serializer
+    {
+        public static byte[] SerialiseBytes(this object obj)
+        {
+            using (var stream = new MemoryStream())
+            {
+                var formatter = new BinaryFormatter();
```
asbjornu: Since `JsonPacket` is serializable to JSON, wouldn't `JsonSerializer` be a better choice than `BinaryFormatter`?

src/app/SharpRaven/RavenClient.cs
```diff
@@ -363,6 +393,19 @@ protected virtual string Send(JsonPacket packet)
             }
             catch (Exception exception)
             {
+                if (!cachePacketsWhenOffline) return HandleException(exception, requester);
+                try
+                {
+                    if (!string.IsNullOrEmpty(packet.Message)) packet.Message += "\n";
+                    packet.Message +=
+                        $"Attempted to send on {DateTime.UtcNow.ToShortDateString()} {DateTime.UtcNow.ToLongTimeString()} UTC however packet failed to send and was cached locally.";
```
asbjornu: I like that you add this info, but I think `Message` should be left untouched. Can you perhaps set this message inside `extra` somewhere instead?

src/app/SharpRaven/Utilities/Serializer.cs
```diff
@@ -0,0 +1,32 @@
...
+using System.IO;
+using System.Linq;
+using System.Runtime.Serialization.Formatters.Binary;
+using System.Text;
+
+namespace SharpRaven.Utilities
+{
+    public static class Serializer
+    {
+        public static byte[] SerialiseBytes(this object obj)
```
asbjornu: Since this is only used for `JsonPacket` and it already does serialization to JSON, I don't think we need a separate class with a general use extension method for `System.Object`.

src/app/SharpRaven/RavenClient.cs
```diff
@@ -106,8 +110,34 @@ public RavenClient(IJsonPacketFactory jsonPacketFactory = null)
             Timeout = TimeSpan.FromSeconds(5);
             this.defaultTags = new Dictionary<string, string>();
             this.breadcrumbs = new CircularBuffer<Breadcrumb>();
+
+            if (cachePacketsWhenOffline) SendCache();
         }
 
+        private void SendCache()
```
asbjornu: I think `SendCachedFiles` is a better, more descriptive name.

asbjornu:
Also, the build is failing on Travis. Can you ensure that all tests are working as well?

src/app/SharpRaven/Data/JsonPacket.cs
samukce: I guess is because is a prime number but I don't know if is really necessary. @rolandh did you get any issue when you didn't use it?

asbjornu: I noticed that this is used as the file name of `JsonPacket` when serialised to disk, which sounds like a bad idea since `GetHashCode` isn't guaranteed to be unique. Using `Guid.NewGuid` or some other way to generate a guaranteed unique, but file-system friendly name, would be better.

src/SharpRaven.sln
rolandh: I think by simply opening and saving the solution it made those changes. I did not modify anything knowingly.

I might not have 3.5 installed, perhaps that is why.

src/app/SharpRaven/Data/Breadcrumb.cs
rolandh: Probably doesn't.

src/app/SharpRaven/Data/JsonPacket.cs
rolandh: I just used a prime number as it is considered good practice. Here is some discussion.

http://stackoverflow.com/questions/3613102/why-use-a-prime-number-in-hashcode
http://stackoverflow.com/questions/263400/what-is-the-best-algorithm-for-an-overridden-system-object-gethashcode

I'm saving a time stamp as well as part of the hashcode, it is true that it could possibly be not unique though especially if you had multiple fail in the same second. This could definitely be improved, I consider it more a proof of concept.

src/app/SharpRaven/Data/JsonPacketFactory.cs
rolandh: By accident probably!

src/app/SharpRaven/Data/SentryEvent.cs
rolandh: I was originally serialising sentryevent, so you are correct it doesn't need to be.

src/app/SharpRaven/RavenClient.cs
rolandh: Sure.

src/app/SharpRaven/RavenClient.cs
rolandh: This could definitely be improved. Do you think a local temporary folder, eg within the executing path? Or use some sort of platform independent temporary folder discovery would be more appropriate?

src/app/SharpRaven/Utilities/Serializer.cs
rolandh: The only reason I did that is I wasn't sure if there was extra information used in the connection side of thigns that was not sent as part of the json packet. Eg does the json serialised packet contain everything that is required to recreate it? Also I was not sure if you had a de-serialise method.

So basically I used an existing method I use in other projects as I was sure it would work and time was limited. If there exists a de-serialise method.

If jsonserialise serialises everything we need, and there is an easy to implement deserialise then yes it would be more appropriate.

src/SharpRaven.sln
rolandh: Regarding you general comments about a pull etc. I am limited for time so I may not get an oppurtunity for some time. So I'm open to other people fixing the discussed issues instead. If no one else is interested then I will attempt it in the coming weeks/months.

Thanks for the comments.

src/app/SharpRaven/RavenClient.cs
rolandh: Can you explain what "Extra" is ? Is it another field accompanied with the message? If so then yes that would definitely be more appropriate.

src/app/SharpRaven/Utilities/Serializer.cs
asbjornu: Since what's sent to Sentry is JSON, it **needs** to contain everything we need. So yes, I believe that's a better way to serialise `JsonPacket` than with `BinaryFormatter`. 😄

rolandh: Is there an existing deserialise method?

asbjornu: @rolandh: No, but as `JsonPacket.ToString()` already does serialisation, it should be easy enough to add something like that. Not sure it belongs in the `JsonPacket` class, but it's not very important, I think.

asbjornu:
@rolandh: Now that the builds are green on both AppVeyor and Travis, do you want to take another jab at this by rebasing it against the `HEAD` of `develop` and addressing the stuff I've pointed out in the review?

bruno-garcia:
This PR is over 1 year old. It has become quite hard to merge due to changes in the codebase.

I'll close it for that reason but feel free to re-open if you'd like to continue on this path.
As @asbjornu mentioned though, it'd be better to split the user injection and the persistence of the event.

Also, I'd rather use json serializer for the persistence instead of binary. There are multiple reasons for that which we can discuss further if you'd like.

rolandh:
It is working OK for our use case. If I ever need to use the latest commit I will add the changes from scratch as that will be easier.

Cheers

PR #181, opened 2017-04-23
Catch exception when creating JsonPacket

### LINKED ISSUE #180: Catch exception during the creation of JsonPacket
The code [here](https://github.com/getsentry/raven-csharp/blob/develop/src/app/SharpRaven/RavenClient.cs#L224) may throw exceptions. I think we should catch it and leave it to `HandleException`.

skyline75489:
This fixes #180 .


samukce:
Hey, good catch. Only one question. When the extra is set a string or integer both are objects, so that issue isn't in the serialization engine? If so, should fixed the serialization is more proactive and the information is not left.

skyline75489:
As for Json.Net, string or integer are not valid `JObject` (JSON object), which is quite understandable. I think here we should make some comments on `Extra` so that people will not misuse it.

samukce:
@skyline75489 yeah, but if we work in the [JsonPacket.Merge](https://github.com/getsentry/raven-csharp/blob/992ffe53ac03827a4242dde72ac3b037faf29f43/src/app/SharpRaven/Data/JsonPacket.cs#L85) method, we can be check if is string or integer primitive and if so we can use **S**tring and **I**nteger class, what you think about this?

More one thing. I saw that you created a test for jsonPacketFactory ([JsonPacketFactoryTests.cs](https://github.com/getsentry/raven-csharp/pull/181/files#diff-da3ac738b60bd7631308ed99111c8825R101)) but you have been changed the code in [RavenClient.cs](https://github.com/getsentry/raven-csharp/pull/181/files#diff-f3090a904e610cbd9b8ca07bc9979bf0R220).  If you change the behavior in the RavenClient class you should create a test form them too. 

Let me know what you think about it.

skyline75489:
Well I'm OK with tweaking `JsonPacket.Merge`. Just that I'm new to sentry and I don't really know what the expected behavior is if you use string or integer or something else for `Extra`.

asbjornu:
@skyline75489: Thanks for this!

The `extra` field is supposed to be a hash map of keys and values that provide more contextual information about the captured event. If someone accidentally sends an `int` or `string` or similar into it, I think we should detect it and handle it gracefully by converting it into a hash map.

The best would be to identify this in a generic way, perhaps by asking JSON.NET whether the object can be serialized to `JObject` and if not wrap it in a dictionary where they key is the data type of the value and the value is what it is.

So basically the `int` value `32` turns into `{ "int": 32 }` and the `string` value `abc` turns into `{ "string": "abc" }`.

Does this sound reasonable and something you could tackle, @skyline75489?

skyline75489:
@asbjornu This looks reasonable. Except that shouldn't we warn users that they are using `Extra` the wrong way? I like the idea of wrapping to dictionary but I don' think something like `{ "int": 32 }` or `{ "string": "abc" }` is what one actually wants.

asbjornu:
@skyline75489: It's difficult to say what the right approach is. But I would rather have the value reported to Sentry and the user discover the error he might have done there, rather than the value not being reported at all and a missed warning logged somewhere the user might not see.

samukce:
I like this asborn approach. Make more sense to me.

Samuel Pereira
Senior Systems Analyst

c. +55 85 9 8883.1368

skype samuel.p.araujo
l. br.linkedin.com/in/samuelpereiraaraujo

g. github.com/samukce

skyline75489:
Alright then. I'll try to come up with something this weekend :)

skyline75489:
Close in favor of #185

Issue #150, opened 2016-09-01
Feature Request: User Feedback


opcon:
Are there any plans to support submitting User Feedback [[1](https://blog.sentry.io/2016/04/21/introducing-user-feedback.html), [2](https://docs.sentry.io/hosted/learn/user-feedback/)] from the C# client? The current implementation for other clients shows a custom webpage, which wouldn't always be appropriate from C#.

I suggest two new methods:
- one that obtains the url for the user feedback webpage currently shown through the PHP and Javascript clients, so the developer can show this to the user if they choose,
- one that allows the developer to submit the user feedback form from C# code

I am happy to start work on this if you can point me to the API endpoints used

samukce:
It's a Good suggestion.

asbjornu:
@opcon Great idea. Since presenting a web page is only one of several possible ways to present a user feedback dialogue in C#, perhaps a good way to do it would be through a `Func<>` of some sort? I haven't looked at what the server-side expects here, but I assume it's just some sort of string, so as long as the implementation returns a string, it can present a dialogue in whatever means natural in the GUI environment, be it a web page, a WPF window, a console application prompt, or whatever.

I'd happily accept a PR for something like this.

opcon:
Awesome. I have a proof-of-concept working, I'll clean it up and submit the pull request.

opcon:
Just an update on this, I've been super busy with Uni stuff recently but I'll be able to get back to this next week

Dids:
@opcon Still waiting. :)

samukce:
@opcon the next week arrived? =P

opcon:
Next week is here! 😆

I submitted a pull-request, #182

asbjornu:
Fixed in #182.

PR #182, opened 2017-04-25
Feature:  Submit User Feedback for an Event

opcon:
This PR adds support for submitting user feedback to Sentry as mentioned in #150.
The feedback form post API endpoint is used, which is undocumented, so
it is possible this may break. However the endpoint has been stable for
the past 7 months or so.

**Notes**:

* The sentry form post API endpoint is found at `https://<sentry url>/api/embed/error-page/?dsn=<dsn>&eventId=<eventId>&name=<name>&email=<email>&comments=<comments>`
* All form fields must be filled out (name, email, comments), and each user feedback must have
an associated eventId
* There is an official API for submitting User Feedback (see [https://docs.sentry.io/api/projects/post-project-user-reports/](https://docs.sentry.io/api/projects/post-project-user-reports/)), however it requires an API key rather than just a Dsn, and so is a little more cumbersome to use.

**Example Usage**:

Here `sentryClient` is a `RavenClient`, and we have already submitted an event and saved the ID in `eventID`.

```
var res = sentryClient.SendUserFeedback(new SentryUserFeedback
    { 
        Name = "Example Name", 
        Email = "Example.Email@null.com", 
        Comments = "This is an example comment.\nMultiline comments should be supported!", 
        EventID = eventID
    });
```
----

**Changes**:

Add the SentryUserFeedback class which holds the feedback information
(Name, Email, Comment, EventID)

Modify the Requester class:

* Add a user feedback specific constructor, which takes a
SentryUserFeedback

* Add the CreateWebRequest method, which takes a URI and returns an
HttpWebRequest object with the boilerplate information setup (timeout,
authorization etc.). This reduces copy-and-pasting code between the two
constructors

* Add the SendFeedback method, which sends the user feedback to Sentry

Modify the Requester.Net45 class:

* Add the SendFeedbackAsync method, which sends the user feedback to
Sentry asynchronously

Modify the RavenClient class:

* Add the SendUserFeedback method, which takes a SentryUserFeedback,
creates the Requester object for the feedback, and sends it to Sentry

Modify the RavenClient.Net45 class:

* Add the SendUserFeedbackAsync method, which does the same job as the
SendUserFeedback method, but asynchronously

Modify the Dsn class:

* Add the FeedbackUri property, which is the user feedback form
submission endpoint

### DIFF
--- a/src/app/SharpRaven/Data/Requester.Net45.cs
+++ b/src/app/SharpRaven/Data/Requester.Net45.cs
@@ -81,6 +81,35 @@ namespace SharpRaven.Data
                 }
             }
         }
+
+        /// <summary>
+        /// Sends the user feedback asynchronously to sentry
+        /// </summary>
+        /// <returns>An empty string if succesful, otherwise the server response</returns>
+        public async Task<string> SendFeedbackAsync()
+        {
+            using (var s = await this.webRequest.GetRequestStreamAsync())
+            {
+                using (var sw = new StreamWriter(s))
+                {
+                    await sw.WriteAsync(feedback.ToString());
+                }
+            }
+            using (var wr = (HttpWebResponse)await this.webRequest.GetResponseAsync())
+            {
+                using (var responseStream = wr.GetResponseStream())
+                {
+                    if (responseStream == null)
+                        return null;
+
+                    using (var sr = new StreamReader(responseStream))
+                    {
+                        var response = await sr.ReadToEndAsync();
+                        return response;
+                    }
+                }
+            }
+        }
     }
 }
 
--- a/src/app/SharpRaven/Data/Requester.cs
+++ b/src/app/SharpRaven/Data/Requester.cs
@@ -49,9 +49,9 @@ namespace SharpRaven.Data
     {
         private readonly RequestData data;
         private readonly JsonPacket packet;
+        private readonly SentryUserFeedback feedback;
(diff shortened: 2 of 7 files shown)


theweavrs:
@asbjornu @opcon what's the status of this PR?

opcon:
@theweavrs as far as I know it's all good on my end, just waiting to hear back from @asbjornu. It's possible this won't be merged because technically there's no official API for submitting user events, and this PR uses API methods deduced from the JavaScript Sentry library.

asbjornu:
I would love to get some input from @dcramer, @mattrobenolt or someone else from @getsentry on this one.

dcramer:
@getsentry/sdks can anyone help provide more context for @asbjornu here?

HazAT:
So I would say the changes are fine, we did something similar for `sentry-cocoa` (but removed it again).

At some point, we should consider providing a proper API for user feedback since sending it like this seems a bit "hacky" and was only built for `raven-js` to be used in browsers.
cc @mitsuhiko we should put user feedback API in our backlog and think of a more general solution for all SDKs providing user feedback.

asbjornu:
@HazAT: Thanks for the input! Is the "proper" user feedback API so far away that this should be merged, or should we wait?

HazAT:
@asbjornu I would say merge ahead!
Since I don't have a timeframe yet when and if we gonna implement the user feedback API.

PR #185, opened 2017-05-13
Convert extra when extra is not serializable

### LINKED ISSUE #180: Catch exception during the creation of JsonPacket

skyline75489:
This fixed #180

### DIFF
--- a/src/app/SharpRaven/Data/JsonPacket.cs
+++ b/src/app/SharpRaven/Data/JsonPacket.cs
@@ -123,7 +123,16 @@ namespace SharpRaven.Data
                 }
             }
             else
-                result = JObject.FromObject(extra);
+            {
+                try
+                {
+                    result = JObject.FromObject(extra);
+                }
+                catch (ArgumentException e)
+                {
+                    result = JObject.Parse(string.Format(@"{{""{0}"":""{1}""}}", extra.GetType(), extra));
+                }
+            }
 
             var jExceptionData = JObject.FromObject(exceptionData);
             result.Merge(jExceptionData);
--- a/src/app/SharpRaven/Data/JsonPacket.cs
+++ b/src/app/SharpRaven/Data/JsonPacket.cs
@@ -128,7 +128,7 @@ namespace SharpRaven.Data
                 {
                     result = JObject.FromObject(extra);
                 }
-                catch (ArgumentException e)
+                catch (ArgumentException)
                 {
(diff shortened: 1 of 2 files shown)


src/app/SharpRaven/Data/JsonPacket.cs
```diff
@@ -123,7 +123,16 @@ private static object Merge(SentryEvent @event)
                 }
             }
             else
-                result = JObject.FromObject(extra);
+            {
+                try
+                {
+                    result = JObject.FromObject(extra);
+                }
+                catch (ArgumentException e)
```
asbjornu: No need for an instance `e` here, is it?

src/app/SharpRaven/Data/JsonPacket.cs
```diff
@@ -123,7 +123,16 @@ private static object Merge(SentryEvent @event)
...
+                {
+                    result = JObject.Parse(string.Format(@"{{""{0}"":""{1}""}}", extra.GetType(), extra));
```
asbjornu: This seems a bit sub-optimal. Can't we instantiate a new `JObject` and give it a property somehow?

src/app/SharpRaven/Data/JsonPacket.cs
skyline75489: Indeed.

src/app/SharpRaven/Data/JsonPacket.cs
skyline75489: It seems the solution is either this or `dynamic` object. I personally do not favor any usage of `dynamic`, though.

asbjornu: Seems like this works just fine:

```c#
result = JObject.FromObject(new Dictionary<string, object>
{
    { extra.GetType().ToString(), extra }
});
```

skyline75489: Yeah there's `Dictionary`... I'm too obsessed with the dynamic type thing in JSON and forgot about it. Should it be `Dictionary<string, string>`, though? Or it's handled by Json.Net itself?

asbjornu: JSON.NET serializes `object` just fine. 😃

asbjornu:
@skyline75489: Excellent work, thanks!

Issue #195, opened 2017-11-16
Sending data to Sentry from custom error aggregator


trurl123:
I want to use raven-sharp from service that gives some information about exception that occured at other servers and applications. So I want to build JsonPacket by myself. Would you please make all proprties at ExceptionFrame as public?

asbjornu:
Sure, pull requests are welcome!

trurl123:
Created https://github.com/getsentry/raven-csharp/pull/197

PR #197, opened 2017-11-22
Add .net standard support

trurl123:
Also make some properties public to use from custom log aggregator
UPDATE: moved properties fix to another PR

### DIFF
--- a/build.cake
+++ b/build.cake
@@ -30,9 +30,9 @@ var version = isTravis ? "0.0.1" : gitVersion.NuGetVersion;
 var artifactsDir = Directory("./artifacts");
 var outputDir = Directory("./build");
 
-var dotnetFrameworks = IsRunningOnWindows() ? new [] { "net45", "net40" } : new string[] { };
+var dotnetFrameworks = IsRunningOnWindows() ? new [] { "net45", "net40", "netstandard2.0" } : new string[] { };
 // net35 can't be build by dotnet - https://github.com/Microsoft/msbuild/issues/1333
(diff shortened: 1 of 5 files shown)


asbjornu:
This PR tackles a bit more than what was discussed in #195. As far as I can see, only 9e026a584d4c5e46f7727e8fd689eaebd6344ccc seems related to that issue. Would it be possible to have that commit (and related changes) in its own PR, so we can discuss .NET Standard support separately?

trurl123:
I removed commits related to https://github.com/getsentry/raven-csharp/issues/195
So this PR contains only .net standard support

beevee:
Looks like 5 out of 10 open pull requests in this repository are directly related to .NET Core or .NET Standard compatibility. Some of them are open for over a year. I'm worried that this PR will share their fate.

What do you guys (that is, maintainers of this repo) think about .NET Core and .NET Standard? Are there any insurmountable techincal difficulties we don't see?

michaeljbaird:
@trurl123 @asbjornu  I just pulled this PR and tested it out on one of my .Net Core Apps 2.0 and it works as expected. I think the Package task in the `build.cake` script needs a little work as it outputs a Nuget package but does not include icon, description...etc. as before but other than that, it should be ready for nuget.org 👍

asbjornu:
@michaeljbaird: That sounds great! I'll merge this and then we can polish the package metadata in a separate pull request. Are you up for the task? 😃

michaeljbaird:
@asbjornu absolutely.

asbjornu:
@michaeljbaird: Excellent. As the .NET 3.5 support was added rather recently, I don't think we're ready to remove it just yet.

PR #198, opened 2017-11-23
Allow sending data to Sentry from custom error aggregator

### LINKED ISSUE #195: Sending data to Sentry from custom error aggregator

trurl123:
Fixes #195


asbjornu:
Thanks!

PR #222, opened 2018-03-07
Add RavenClientNoop, an empty (no-op) empty implementation of IRavenClient

### LINKED ISSUE #213: Noop implementation of IRavenClient
I think it would be nice to have a no-op implementation of IRavenClient available as that helps with dependency injection on development builds where I don't want to have nulls injected but rather an implementation that does not send data anywhere.

Would it make sense to include it in main distribution?

I can make a PR for this, however I am not sure about the naming of the class.

razzmatazz:
Fixes #213.

Note, that I am not sure about the value of RavenClientNoop.Dsn. Currently, I have it set to:

```c#
            currentDsn = new Dsn("http://localhost/no-op-dsn");
```

Which I am not sure is a right thing to do.

### DIFF
--- a/src/app/SharpRaven/NoOpRavenClient.Net45.cs
+++ b/src/app/SharpRaven/NoOpRavenClient.Net45.cs
@@ -1,0 +1,113 @@
+#region License
+
+// Copyright (c) 2014 The Sentry Team and individual contributors.
+// All rights reserved.
+//
+// Redistribution and use in source and binary forms, with or without modification, are permitted
+// provided that the following conditions are met:
+//
+//     1. Redistributions of source code must retain the above copyright notice, this list of
+//        conditions and the following disclaimer.
+//
+//     2. Redistributions in binary form must reproduce the above copyright notice, this list of
+//        conditions and the following disclaimer in the documentation and/or other materials
+//        provided with the distribution.
+//
+//     3. Neither the name of the Sentry nor the names of its contributors may be used to
+//        endorse or promote products derived from this software without specific prior written
+//        permission.
+//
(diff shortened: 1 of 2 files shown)


src/app/SharpRaven/RavenClientNoop.Net45.cs
```diff
@@ -0,0 +1,113 @@
...
+
+using SharpRaven.Data;
+
+namespace SharpRaven
+{
+    /// <summary>
+    /// Empty (no-op) implementation for the Raven Client for use in dependency injection
+    /// and other places when a silent operation is needed.
+    /// </summary>
+    public partial class RavenClientNoop
```
asbjornu: Can we rename this to `NopRavenClient`, `NoOpRavenClient` or `NoopRavenClient`?

src/app/SharpRaven/RavenClientNoop.cs
```diff
@@ -0,0 +1,239 @@
...
+            return Guid.NewGuid().ToString("n");
+        }
+
+
+        /// <summary>Sends the specified user feedback to Sentry</summary>
+        /// <returns>An empty string if succesful, otherwise the server response</returns>
+        /// <param name="feedback">The user feedback to send</param>
+        public string SendUserFeedback(SentryUserFeedback feedback)
+        {
+            return "";
```
asbjornu: Can you change this to `string.Empty`?

asbjornu:
I think the DSN is fine, but if you want, you can change it to the [reserved `invalid.` host name as per RFC 6761](https://tools.ietf.org/html/rfc6761#section-6.4)

razzmatazz:
Thanks for the review.
I have now rebased the changes with class renamed to NoOpRavenClient.

asbjornu:
Great, thanks! A pre-release containing this change should appear on NuGet.org in a few minutes.

razzmatazz:
Great!

PR #225, opened 2018-03-11
Set configuration to release in AppVeyor

asbjornu:
The `Configuration` wasn't set to `Release` when building on AppVeyor, so the produced NuGet package is built in `Debug` mode now.


asbjornu:
Wrong. Seems like the `Configuration` is set to `Release` explicitly in the Cake build script. Closing.

PR #231, opened 2018-03-26
Fixed NoOpRavenClient dsn

### LINKED ISSUE #230: Exception when instantiating NoOpRavenClient
Instantiating a NoOpRaventClient will result in a Invalid Dsn exception.

here's the trace
```
EXCEPTION: System.ArgumentException: Invalid DSN
Parameter name: dsn ---> System.IndexOutOfRangeException: Index was outside the bounds of the array.
   at SharpRaven.Dsn.GetPrivateKey(Uri uri)
   at SharpRaven.Dsn..ctor(String dsn)
   --- End of inner exception stack trace ---
   at SharpRaven.Dsn..ctor(String dsn)
   at SharpRaven.NoOpRavenClient..ctor()

```

I tried to pass the `http://localhost/no-op-dsn` and `http://sentry-dsn.invalid` both are considered invalid. The issue is obviously from the Dsn class. 

_building for netcoreapp2.0_

alirezajm:
Changed NoOpRavenClient DSN from `http://sentry-dsn.invalid to` `http://ig:nore@sentry-dsn.invalid` so the Dns class constructor won't throw an exception, originating from the below from Dns.cs. 

```c#
private static string GetPrivateKey(Uri uri)
        {
            return uri.UserInfo.Split(':')[1];
        }
```

Although that also can throw a more meaningful exception.

Fixes #230.

### DIFF
--- a/src/app/SharpRaven/NoOpRavenClient.cs
+++ b/src/app/SharpRaven/NoOpRavenClient.cs
@@ -49,7 +49,7 @@ namespace SharpRaven
         /// </summary>
         public NoOpRavenClient()
         {
-            currentDsn = new Dsn("http://sentry-dsn.invalid");
+            currentDsn = new Dsn("http://ig:nore@sentry-dsn.invalid");
             defaultTags = new Dictionary<string, string>();
         }
 


razzmatazz:
I can confirm this is broken in my original PR #222 as I did not properly test it that time. And now that I have attempted to use the NoOpRavenClient from nuget package I discovered that I uploaded broken code to that PR.

My patch to fix it was:
```
--- a/src/app/SharpRaven/NoOpRavenClient.cs
+++ b/src/app/SharpRaven/NoOpRavenClient.cs
@@ -49,7 +49,7 @@ public partial class NoOpRavenClient : IRavenClient
         /// </summary>
         public NoOpRavenClient()
         {
-            currentDsn = new Dsn("http://sentry-dsn.invalid");
+            currentDsn = new Dsn("http://00000000000000000000000000000000:00000000000000000000000000000000@sentry-dsn.invalid/1");
             defaultTags = new Dictionary<string, string>();
         }
 ```

But your changes in this PR are probably good too if you manage to instantiate `NoOpRavenClient` instance.

Stand-alone unit tests for `raven-csharp` would be nice to have..

awbacker:
How long until 3.0?  Its august now, and I need to manually copy the NoOpRaven class over to my solution to get his running.

bruno-garcia:
We've been busy with the new unified API in the [new SDK](https://github.com/getsentry/sentry-dotnet).

Can you use the latest pre-release? If we released 3.0 right now that's exactly what you would be running and it'd be great to know it doesn't break people before releasing it.

PR #232, opened 2018-04-05
Private key is not longer required for DSN

bruno-garcia:
The private DSNs (including secret keys) will be officially deprecated soon. This PR allows users to specify a public DSN and makes sure the sentry_secret is not included in the header anymore. For backwards compatibility, a private DSN can still be used, though.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -30,7 +30,7 @@ Instantiate the client with your DSN:
 
 .. sourcecode:: csharp
 
-    var ravenClient = new RavenClient("___DSN___");
+    var ravenClient = new RavenClient("___PUBLIC_DSN___");
 
 Capturing Exceptions
 --------------------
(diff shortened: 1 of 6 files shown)


bruno-garcia:
Travis has been broken for a while. Let's fixed in on a separate PR.
I'll just see if @asbjornu can take a look at the changes and we can merge.

docs/index.rst
```diff
@@ -106,7 +106,7 @@ The only thing you have to do is provide a DSN, either by registering an instanc
 
     protected override void ApplicationStartup(TinyIoCContainer container, IPipelines pipelines)
     {
-        container.Register(new Dsn("http://public:secret@example.com/project-id"));
+        container.Register(new Dsn("http://public@example.com/project-id"));
```
jan-auer: *nit*: You could use `___PUBLIC_DSN___` here to show an interactive DSN chooser in docs and the setup wizard:

```csharp
container.Register(new Dsn("___PUBLIC_DSN___"));
```

bruno-garcia: Done!

mogikanin:
The question about dsn w/o private key. I tried to use dsn with only public key, but our sentry server doesn't accept such requests and return 403 error. Is it should be somehow configured on server side as well, or probably it's needed specific sentry version? We use 8.19.0

jan-auer:
The private DSN is still required up to Sentry 8.22.0. The release of Sentry 9 to on-premise customers is just around the corner and will fix this.

mogikanin:
ok, thanks @jan-auer

PR #237, opened 2018-04-18
feat: Sentry context

bruno-garcia:
* Payload creating has runtime detection
* Context is built lazily on first event
* Root context is closed per event to avoid sys calls required to build it
* .NET Framework 4.7.1 added to support NET Standard 2.0 paths (RuntimeInformation)
* Warnings over obsolete calls removed from test projects
* Nancy targets net471 also
* Json.NET indirect reference to Nancy


src/app/SharpRaven/Data/Context/App.cs
```diff
@@ -0,0 +1,147 @@
...
+                var app = new App();
+
+                var asm = ApplicationName ?? EntryAssemblyNameLocator.GetAssemblyName();
+                if (asm != null)
+                {
+                    app.Name = asm.Name;
+                    app.Version = asm.Version?.ToString();
+                }
+
+                using (var proc = Process.GetCurrentProcess())
```
asbjornu: Is it wise to `.Dispose()` the current process?

src/app/SharpRaven/Data/Context/Device.cs
```diff
@@ -0,0 +1,220 @@
...
+                SystemUtil.WriteError(e);
+                return null;
+            }
+        }
+
+        /// <summary>
+        /// Retrieves the architecture of this device.
+        /// </summary>
+        /// <returns>Arch or null if unknown</returns>
+        internal static string GetArchitecture() =>
```
asbjornu: All these `internal` methods only seem to be used in a `private` scope, so can they be changed to `private` instead?

src/app/SharpRaven/Data/SentryEvent.cs
```diff
@@ -68,8 +70,15 @@ private SentryEvent()
...
             Tags = new Dictionary<string, string>();
             Fingerprint = new List<string>();
+            this.contexts = Contexts.Create();
         }
 
+        /// <summary>Represents Sentry's structured Context</summary>
+        /// <seealso href="https://docs.sentry.io/clientdev/interfaces/contexts/"/>
+        public Contexts Contexts
+        {
+            get { return this.contexts; }
```
asbjornu: Can't you just add an `internal set;` to this property so we get rid of the field?

src/app/SharpRaven/Utilities/EntryAssemblyNameLocator.cs
```diff
@@ -0,0 +1,86 @@
...
+            return entryAssemblyName;
+        }
+
+        internal static AssemblyName GetAssemblyNameFromEntryMethod()
+        {
+            var entryMethod = GetApplicationEntryMethod();
+            return entryMethod?.Module?.Assembly?.GetName();
+        }
+
+        internal static MethodBase GetApplicationEntryMethod()
```
asbjornu: 👍

src/app/SharpRaven/Utilities/RuntimeInfoHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+            string monoVersion = null;
+
+            var type = Type.GetType("Mono.Runtime", false);
+            if (type != null)
+            {
+                var displayName = type.GetMethod("GetDisplayName", BindingFlags.NonPublic | BindingFlags.Static);
+                if (displayName != null)
+                {
+                    monoVersion = displayName.Invoke(null, null) as string;
+                    if (monoVersion != null)
```
asbjornu: Can't all of these `null` checks be collapsed into a chained `null` propagation (`?.`) expression?

src/app/SharpRaven/Utilities/RuntimeInfoHelper.cs
```diff
@@ -0,0 +1,100 @@
...
+                        ? $".NET Framework {CheckFor45PlusVersion(releaseId)}"
+                        : null;
+                }
+            }
+            catch (PlatformNotSupportedException)
+            {
+                return null;
+            }
+
+            string CheckFor45PlusVersion(int releaseKey)
```
asbjornu: A bit unintuitive that a method called `Check...()` returns `string`.

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+namespace SharpRaven.UnitTests.Data.Context
+{
+    [TestFixture]
+    public class AppTests
+    {
+        [Test]
+        public void Create_StartTime_SometimeInThePast()
+        {
+            var sut = App.Create();
```
asbjornu: I'm not a fan of `sut`. Can't this just be called `app`?

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            Assert.True(sut.StartTime.HasValue);
```
asbjornu: I think `Assert.That(app.StartTime.HasValue, Is.True)` reads better.

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            Assert.AreNotEqual(default(DateTimeOffset), sut.StartTime.Value);
```
asbjornu: I think `Assert.That(app.StartTime.Value, Is.Not.EqualTo(default(DateTimeOffset))` reads better.

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            Assert.Greater(DateTimeOffset.UtcNow, sut.StartTime);
```
asbjornu: I think `Assert.That(app.StartTime, Is.GreaterThan(DateTimeOffset.UtcNow))` reads better.

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+#if !NET35
+        [Test]
+        public void Create_Version_EqualEntryAssembly()
+        {
+            var sut = App.Create();
+
+            Assert.NotNull(sut.Version);
```
asbjornu: I think `Assert.That(app.Version, Is.Not.Null)` reads better.

src/tests/SharpRaven.UnitTests/Data/Context/AppTests.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Theory]
+        [Test, TestCaseSource(typeof(AppTests), nameof(TestCases))]
+        public void SerializeObject_TestCase_SerializesAsExpected(TestCase @case)
+        {
+            var actual = JsonConvert.SerializeObject(@case.Object);
+
+            Assert.That(actual, Is.EqualTo(@case.ExpectedSerializationOutput));
+        }
+
+        public static IEnumerable<object[]> TestCases()
```
asbjornu: 👍

src/app/SharpRaven/Data/Context/App.cs
bruno-garcia: Yes. This instance was created by me at this location and should be [disposed when not needed](https://stackoverflow.com/questions/28023490/should-one-call-dispose-for-process-getcurrentprocess).

src/app/SharpRaven/Data/Context/Device.cs
bruno-garcia: Yes, I made them internal for testing but I'm just covering them through testing the whole Device at the end. I'll make them private.

src/app/SharpRaven/Utilities/RuntimeInfoHelper.cs
bruno-garcia: I was afraid it'd be less readable but it's fine. Will change.

src/app/SharpRaven/Utilities/RuntimeInfoHelper.cs
bruno-garcia: True. I'll make it: Get

src/app/SharpRaven/Data/Context/Device.cs
bruno-garcia: It is actually used by tests*

bruno-garcia:
@asbjornu could you please check again?

asbjornu:
@bruno-garcia: `GetArchitecture` and `GetApplicationEntryMethod` are still `internal`, but otherwise I think it looks good.

bruno-garcia:
@asbjornu I commented on the review: It's to be accessed by the Unit testing library.

bruno-garcia:
@asbjornu if that's ok, could you please approve it? Thanks!

PR #238, opened 2018-04-20
Tests pass on macOS

bruno-garcia:
* net45 takes new dependency instead of relying on Win Registry
* TimeZoneInfo Local on tests to bypass Win/Unix string id not mapped
* AppName found on Mono looking at InternalInvoke method
* CPU arch takes Mono approach as fallback (Environment.Is64bit)

### DIFF
--- a/src/app/SharpRaven/Data/Context/Device.cs
+++ b/src/app/SharpRaven/Data/Context/Device.cs
@@ -30,6 +30,9 @@
 
 using System;
 using System.Diagnostics;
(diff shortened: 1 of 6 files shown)


asbjornu:
Why has Travis stopped reporting build status?

bruno-garcia:
I've disabled as the build was broken for over a month. 
I noticed you commented on the cake build issue. 
I'll try to investigate the build (I work on a Mac so I can test it locally before pushing to Travis.

PR #247, opened 2018-04-30
fix(device): Report Device architecture or Null if it cannot reliably

### LINKED ISSUE #245: Device architecture reported based on process architecture
When targeting .NET Framework 4+ the API used to extract CPU architecture uses the [bitness of the process instead of the machine](https://github.com/getsentry/raven-csharp/blob/4df3d8844ef8009752a66fd3793ac8feaccb9141/src/app/SharpRaven/Data/Context/Device.cs#L220).

Contexts interface for Device architecture expects the bitness of the actual device.

* Consider adding application bitness `Environment.Is64BitProcess` to App Contexts.
* Remove the use of `Is64BitProcess` from `Device`

bruno-garcia:
Resolves #245


src/app/SharpRaven/Data/Context/Device.cs
```diff
@@ -212,17 +213,46 @@ internal static Device Create()
...
+                    }
+                }
+            }
+
+            return !is64Bit.HasValue ? null : is64Bit.Value ? "X64" : "X86";
+        }
+
+        [DllImport("kernel32.dll", CallingConvention = CallingConvention.Winapi)]
+        [return: MarshalAs(UnmanagedType.Bool)]
+        private static extern bool IsWow64Process([In] IntPtr hProcess, [Out] out bool wow64Process);
```
asbjornu: Will this just silently fail and return `null` on Mono?

bruno-garcia: On Windows it should work. On non-Windows it returns `false`.

P/Invoke calls return `false` if they fail. We can set `SetLastError = true` which allows us to get it via `Marshal.GetLastWin32Error` but I thought it's not worth making another platform call to find out why it was not possible to get the value. If it fails to get it, we return null and don't report any Arch.

PR #257, opened 2018-06-07
fix: report device arch instead of process arch

bruno-garcia:


bruno-garcia:
True though if the CPU is 64 but your OS is 32 bit, not sure how useful this info is or what kind of API there is to retrieve this info.

Issue #260, opened 2018-06-13
No stack traces for intentional error captures


dcc635:
None of these calls show a stack trace in Sentry:

- `ravenClient.Capture(new SentryEvent(message));`
  - shows the message in Sentry but no stack trace
- `ravenClient.Capture(new SentryEvent(new Exception(message)));`
  - shows the message in Sentry but no stack trace
- `ravenClient.Capture(new SentryEvent(new Exception()));`
  - shows `Exception of type 'System.Exception' was thrown.` in Sentry but no stack trace

Uncaught exceptions show the full stack trace. It'd be nice if these intentional error reports also showed stack traces. My workaround now is to make the messages unique so I can grep where they occur.

bruno-garcia:
Hi @dcc635, that's a fair request. The reason its currently not there when there is no Exception is that the stacktrace is currently pulled out of the exception. 

Could you please describe what kind of app you are building? 
Also, what is the target framework?

We might be able to solve that with `Environment.Stacktrace` or something

jviolas:
Hi ! 
I had the same issue today.
I've search a workaround without touching to SharpRaven code but it was not ideal.

I've managed to make the changes and I'll submit a PR.

@bruno-garcia For info, in my case, it's in the context of a desktop WPF application targeting .net 4.5.2
And I was not able to use `Environment.Stacktrace` because it's a string. I used directly the `StackTrace` class.

bruno-garcia:
Hi @jviolas. Looks good to me but better ask @asbjornu when the PR comes :)

PR #264, opened 2018-07-11
Add possibility to force a StackTrace capture when no frames found

jviolas:
When capturing an Exception in the same code that create it, no stacktrace is sent to Sentry.

See #260 for details.

This PR adds an option when an Exception is captured to force Raven to generate the stacktrace.

I've only added this option for `Exception` through `new SentryEvent(Exception exception, bool forceCaptureStackTrace = false)`. It could be added for messages but it would require to either change `SentryMessage` to have a possible StackTrace or consider those messages as `SentryException`


asbjornu (approved):

jviolas:
@bruno-garcia I commited changes from your suggestion. '
Just dont really know why it's not building. It seems the tests I've added are failing but they work when I launch it locally. From the logs, it seems the "internal stack frame skipping" is not working when launched in the context of the build. Any idea why ?

src/app/SharpRaven/Data/SentryStacktrace.cs
```diff
@@ -46,18 +46,26 @@ public class SentryStacktrace
...
             var trace = new StackTrace(exception, true);
+            
             var frames = trace.GetFrames();
 
-            if (frames == null)
+            if (frames == null && forceCaptureStackTrace)
+            {
+                // Skip first frame if called from this assembly (in case of force capture stacktrace)
+                frames = new StackTrace(true).GetFrames().SkipWhile(f => f.GetMethod().Module.Name == "SharpRaven.dll").ToArray();
+            }
```
bruno-garcia: One last thing: I suggest using `f.GetMethod()?.Module.Name == "SharpRaven"` here, unless I'm mistaken GetMethod can return null.

bruno-garcia:
Honestly I'm not sure.
On Travis, it runs on Mono so there might be a surprise there.
You can check if it's running on Mono and skip your test. [Something like this](https://github.com/getsentry/dotnet-sentry-platform-abstractions/blob/dev/test/Sentry.PlatformAbstractions.Tests/RuntimeInfoTests.cs#L48). You can easily check with ` Type.GetType("Mono.Runtime", false) != null`

On Windows it should behave the same, under .NET Framework. Running the cake build on your machine works and on appveyor doesn't is very strange.

jviolas:
It seems the StackTrace is empty on Mono. I'll try to debug this later.

jviolas:
Yeah ! 🎉 Found it. So in Mono, `StackTrace.GetFrames()` returns an empty array instead of null when there is no frame. So my checks in `SentryStacktrace` constructor was not correct.

bruno-garcia (approved):
Looks good! Thanks

PR #270, opened 2018-08-30
Use X-Forwarded-For header as client IP address when possible

olsh:
When IIS is hosted behind proxy server like nginx, the user field contains the IP address of the nginx server, not actual client IP address.
The code tries to get the actual client IP from `X-Forwarded-For` header if possible.


src/app/SharpRaven/Data/SentryUserFactory.cs
```diff
@@ -85,7 +86,8 @@ private static dynamic GetIpAddress()
         {
             try
             {
-                return SentryRequestFactory.HttpContext.Request.UserHostAddress;
+                return SentryRequestFactory.HttpContext.Request.Headers.GetValues("X-Forwarded-For")?.FirstOrDefault()
```
nokitakaze, line 89: It shouldn't be like that. The client could just send this header and he will hide his real IP.
It should be with some boolean property like `useXForwardedForHeaderIfExist` **with default value `false`**

olsh: Well, I thought that is not a critical issue for Sentry events from a security standpoint.
You can always get the `real IP` in Headers.
The `useXForwardedForHeaderIfExist` option also is not a perfect solution.
The problem can be solved with a whitelist of reverse proxy IPs, but as I said before, this is not a critical issue.

